PR #5, opened 2018-03-03
Lightning input fsc

ryemcc:
I thought the Lightning Input component would give a lot of bang for the buck, since changing the type attribute gives so much flexibility. Let me know what you think.


flow_screen_components/LightningInputFSC/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+* type: "number"
+   * label: "Enter a dollar amount" 
+   * formatter: "currency" 
+   * step: "0.01" 
+
+* type: "password"
+   * label: "Password" 
+
+* type: "radio"
```
alexed1: their implementation of radio here seems a little dysfunctional. first of all radio buttons have always been single select. Here, you could select several at once. But that's what checkboxes are for. And you have to write code to look at each once, check the value of checked, and take appropriate action. I can't think of a scenario where I'd prefer this to the built in radio button solution in Flow (which is not without its own downsides). Leave it in, I guess, but I'm inclined to deemphasize it...

flow_screen_components/LightningInputFSC/README.md
```diff
@@ -0,0 +1,126 @@
...
+   * label: "Enter a percentage value" 
+   * formatter: "percent" 
+   * step: "0.01" 
+
+* type: "number"
+   * label: "Enter a dollar amount" 
+   * formatter: "currency" 
+   * step: "0.01" 
+
+* type: "password"
```
alexed1: will probably deemphasize the ones that don't seem to add any value over the built-in components in Flow and play up the ones that are new. But you might be able to educate me on things these can do that Flow can't do out of the box. If there are such distinctions, let's highlight them here in the docs.

flow_screen_components/LightningInputFSC/README.md
```diff
@@ -0,0 +1,126 @@
+# Lightning Input
+This is a port of the [Lightning Input](https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/aura_compref_lightning_input.htm) component, which includes the following input types:
```
alexed1: I wouldn't say "port". I'm generally going with "a Flow-optimized version of"

flow_screen_components/LightningInputFSC/README.md
```diff
@@ -0,0 +1,126 @@
...
+* type: "url"
+   * label: "Website" 
+
+* type: "week"
+   * label: "Week" 
+
+## Custom Validation Parameters and Messages
+
+### The following validations can be set in the Flow to show default or custom messages:
+formatter | max | maxlength | min | minlength | required
```
alexed1: You've added a lot of richness here, and some of it needs a little more doc. For example, for "formatter", it would be great to show the three legit values and give an example.

flow_screen_components/LightningInputFSC/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+checkbox | checkbox-button | color | date | datetime-local | email | month | number | password | radio | range | tel | time | toggle | url | week
+
+## How It Works ##
+
+The component has two required attributs: `type` (possible values noted above) and `label`, which must be set in the Flow. There are also several type-specific attributes and validation attributes available ([see here for full documentation]((https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/aura_compref_lightning_input.htm)))
+
+A `value` attribute captures the user's input and can be used to set a Flow variable.
+
+For checkboxes and radio buttons, setting the "Group Name" attribute in the Flow will tie the buttons together.
```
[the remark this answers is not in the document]
ryemcc: Yep, that's exactly how it works. I'll clarify in the docs.

flow_screen_components/LightningInputFSC/README.md
ryemcc: I was confused by the multiselect radios too. I've noticed multiselect radio buttons are used all over the on the SLDS site (e.g., the facets to the right of their component examples)--seems like a design choice being made there.

alexed1: Sigh. Kids these days.....

PR #8, opened 2018-03-13
Add WhereClause Support

datapharmer:
Supports crude SOQL where queries to provide basic filtering capabilities. Input is not sanitized.


datapharmer:
whereclause

Sorry for the delay. Let me know if this works for you!

PR #17, opened 2018-04-24
Dynamic picklist

ryemcc:
Hi Alex ,

Here is a dynamic picklist component. It grabs the values for a given object/field using a getDescribe call and builds a picklist for flow.

Best,
Ryan
@alexed1


alexed1:
this is awesome.

alexed1:
Question for you: how do you see this different from/improving from the native capability of flow to create a drop down list box that points at a picklist? 

screen shot 2018-05-07 at 2 03 09 pm

ryemcc:
There is overlap, but it saves clicking through some dialogue screens, assigning unique names, etc. when you know the field you want to grab.

PR #43, opened 2018-06-19
Adds support for Salesforce Classic

datapharmer:
Adds detection and support for Salesforce Classic under select scenarios. Function under lightning appears to be broken but is annotated. Additional console logging has been added.


flow_screen_components/navigationButtonFSC/force-app/main/default/aura/navigationButtonFSC/navigationButtonFSC.cmp
```diff
@@ -1,9 +1,11 @@
 <aura:component implements="flexipage:availableForAllPageTypes,lightning:availableForFlowScreens" access="global">
```
alexed1: I didn't know this stuff could be made to work in Classic. That's very cool.

Can you review my responses to the problems you were having getting the navigation to work in Lightning? I want to know if that changes anything you want to do with this pull request. 

Can you point me at an org that has this version of the component installed, so I can see it in action?

datapharmer: Yes, this is now setup and you can test this in the same sandbox I gave access to before. 

The only issue I'm seeing with this merge is that based on the quick actions/custom button info you provided the detection is still returning that it is the lightning environment when using the custom button, so any suggestions on how to fix that detection scenario would be appreciated (you can see the detection if you run the developer console in your browser). 

I'm also still looking for a way to do a URLFOR so I can obtain a  CSRF confirmation token but seem to be at a dead end there... any documentation or suggestions on how to get the CSRF token into the url from a flow would be sincerely appreciated.

alexed1:
nice work.

PR #52, opened 2018-07-05
Update lookupFSC.design

ericrsmith35:


ericrsmith35:
Part of Default Value update for lookupFSC

PR #53, opened 2018-07-05
Update lookupFSC.cmp

ericrsmith35:
Part of Default Value update for lookupFSC


alexed1:
Hi, Eric. Can you delete all of these individual pull requests and then create a single pull request that reflects all of the differences between your branch and master? That's the best practice.... 

The reason this is ending up complicated is that you're creating too many branches. You don't want to be creating a branch for each file. You want to create a single "working" branch that you make all of your changes to. It's fine to branch from that but you want to keep merging your changes into your working branch so it reflects all of your work. Then when it comes time to create your pull request, git will generate one that shows all the differences between your working branch and master branch.

If you make commits frequently then you don't have to worry that a mistake in file X will cause you to lose work in file Y, because you can always go back through your commit history to right before the problem and branch from there.

PR #68, opened 2018-08-11
Ericbranch1

ericrsmith35:
Thanks to some assistance from John Pipkin we can now group multiple separate Parent/Child components on a single flow screen.


alexed1:
nice!

I don't have time right now to properly support cool adds like this. Write up a blog post describing what you've added and I'll publish it on the blog I use for flow stuff.

PR #94, opened 2018-11-29
imageButtonfsc Update

ericrsmith35:
Added the option to add up to two additional images
One is for when the mouse is hovered over the image
The other for when the mouse button is pressed while over the image
		
This allows for 3-phase or 3D type image buttons to be supported


ericrsmith35:
Thanks.  I have updated the What's New page, the Image page and the Image -
Documentation page.

I also added a new cartoon to the Home page.

Eric

PR #96, opened 2018-11-30
Update for navigationButtonFSC

ericrsmith35:
I have updated this component to correctly handle Classic vs Lightning with the addition of an APEX Class.

I have enhanced the component to now support two buttons.  This will allow for a cleaner OK/Cancel type logic on a flow screen.


ericrsmith35:
Thanks Alex.  I've updated the Unofficial Flow site for this component
enhancement.

PR #122, opened 2019-01-29
Apply default value

chexxor:
When a value is passed into the component from the `value` attribute, select it as the default value in the option list that is returned from the server.

This gets around the limitation of the built-in picklist input in flows. The workaround described in the following blog post shows the selected option twice:
http://www.sfdcr.com/automation/visual-flow-series-default-current-picklist-value/


alexed1:
thanks for submitting! I've asked Ryan, the original author of this component, to review your pull request so we can merge it in.

chexxor:
I have a pretty busy life, so I'm not sure when I'll be able to come back to make changes and review this. Until then I wanted to take the time to say I appreciate your review, and that this is on my radar. 👍

PR #127, opened 2019-02-22
Update LightningLookupHelper.js

ericrsmith35:
Fixed issue where a value with an apostrophe in it would crash the SOQL lookup


alexed1:
Eric, do you think you could create a new package for installing this component? If you can provide me with an install link, I'll update the page at unofficialsf.com

ericrsmith35:
https://login.salesforce.com/packaging/installPackage.apexp?p0=04t1K000002ijQU&isdtp=p1

PR #132, opened 2019-03-07
Lookup fsc bug fix

ericrsmith35:
Remove jQuery references to avoid issues with Lightning Locker Service


alexed1:
Great. pls create a new package for unofficialsf.com. Make sure to update the page at unofficialsf.com and not the old page at the google site

PR #134, opened 2019-03-10
Eric datatable

ericrsmith35:
This is a major enhancement to the datatableFSC component.  It has been updated to offer more features and to be more generic and easily customized.

Features:
- Up to 10 columns with only predefined columns showing
- Parameters for Column Width and Column Alignment
- All Columns are sortable
- An Icon can be displayed with the first Column Header
- Support built in for up to 8 Standard and Custom Objects
- Already configured to support Account, Asset, Case, Contact, Contract, and Opportunity


alexed1:
this is great. documentation will be key here.

PR #144, opened 2019-03-19
Validation function lookup fsc

ludocal:
As Winter '19 came with a validation function for custom screen component.
I update the lookup to manage this new feature.


alexed1:
This is great!

@ericrsmith35 , when you get a chance, can you pull this code to your org and test it? Then I'll merge it in.

ericrsmith35:
@alexed1 Test passes fine.  Ok to merge.

alexed1:
Ludovic, would you like to generate a new package of this component so we can post an updated install link that includes your validation logic? The org doesn't really matter much: you can use a dev edition org. make sure the apex tests get added to the package. Eric knows the drill and can also help if you get stuck. I'll take the install link and put it here: https://unofficialsf.com/lookup/

PR #154, opened 2019-04-19
DatatableFSC v2.4

ericrsmith35:
Updated to support Pre-Selected Rows, No Row Selection, Single Row Selection (Radio Buttons), Multi-Row selection with a row limit.


ericrsmith35:
I've already updated unofficialsf.com

PR #168, opened 2019-06-11
Evan dual list box enhancements

RealEvanPonter:
I've added in some additional attributes to the dualListBox component that helped with my implementation.


RealEvanPonter:
Hi Alex,

I've added in the validation to ensure that a value is present when the required attribute is set to true. Let me know if any other changes are needed.

-Evan

alexed1:
Great. Can you update the documentation at https://wordpress.com/block-editor/post/unofficialsf.com/12801 ? I'll give you editor access....

RealEvanPonter:
All set. I'm not sure how to update the install links for the
managed/unmanaged packages to use the latest source code - I've never done
that before

PR #169, opened 2019-06-11
Evan minimize utility item

RealEvanPonter:
I created a component that will minimize the utility bar item. This is used at the end of flows that are embedded in the utility bar so that they close instead of starting over. If you also use the same flow in non-utility bar configurations, this component will not throw any errors.


ericrsmith35:
@RealEvanPonter I like this.  Very nice utility.

RealEvanPonter:
Sure, let me know what the credentials are and I can create the managed package. I'll email you the blog post and a screencast.

Thanks!

PR #173, opened 2019-07-01
Lookup fixes

ericrsmith35:
Fix error handling for Required attribute

Escape out the ' character from the generated SOQL statement that does the lookup


ericrsmith35:
I've already updated unofficial except I don't remember which org the
Managed Package was created in.  Is there any way to find out that you know
of?

Eric

Issue #180, opened 2019-09-16
NavigateToSObject improvement request


dleland:
Hello - NavigateToSObject works great and I use it a lot.  There is one thing I wish it had - the ability to be either in "view" or "edit" mode when navigating to the object.  Currently, it's always in "view".

Thanks!

michaeldawson5:
Hello - we are using the sales console and when the flow is finished it opens a new tab but not in the subtab that you are working in, it would be great if it opened in the subtab.

PR #184, opened 2019-09-19
Ers navigate to record

ericrsmith35:
This is a generic Lightning Flow Action Component that can be used at the end of a Flow to load a Salesforce Record page. Unlike the Navigate to SObject component, you can specify whether to open the record in Edit or View mode.

I wrote this component to solve the issue raised in this idea:

NavigateToSObject improvement request (#180)

This component uses lightning:navigation rather than the deprecated force:navigateToSObject that is in the Navigate to SObject component.


ericrsmith35:
I renamed the component but I'm not sure how to rename the directory.

Eric

ericrsmith35:
Alex - this should be ready to go.

rraocranevscom:
Am not sure why it is NOT working for me. I have the navigatetorecord(also tried naviageteeverywhere) as the last action element in the flow with specifying the new created record id, object (Account) and the view (also edit) mode. Stored ouput variable. When the flow ends i get the flow name and beneath says Flow completed successfully. 
I need it to go to the newly created record for which i have given the record id and object name.
Can somebody help me with this?? Not sure where i am going wrong and have spent much time trying to figure it out.

ericrsmith35:
Go ahead and post an image of your flow and some screenshots of how the navigate component is configured.

alexed1:
Navigation problems almost always are because you're running it in one of
the places that don't work with force navigation events. Clicking run from
the Flow Builder, for example, creates a runtime session that doesn't work
with navigation events. How are you launching this flow?

alex

Alex Edelstein
Sr. Dir. Product Management - Automation and Decisioning
Salesforce

<http://smart.salesforce.com/sig/aedelstein//us_mb/default/link.html>

Ani26:
Hi!

I have a problem, the component does not appear in the left listing of the flow canvas. What will I have wrong?

ericrsmith35:
The component is a Flow Action.  Drag the Action component onto the canvas
and then search for 'navigate'

Collins1313:
Hi , the component works for me then run from the object page using  custom button, however when i try and use this same flow run from a list button on a related list. the navigate action just seems to reload the flow. Have you got any ideas why this would be the case? Is it not possible to use this when starting a flow via a related list button and do you have any ideas of what i could do instead?

ericrsmith35:
@Alex Edelstein <aedelstein@salesforce.com> is there a writeup somewhere on
how Flow Actions behave based on how they are initiated?

PR #199, opened 2019-10-30
Horiz rule

ericrsmith35:
Updated horizontalRuleFSC to support placement on Lightning pages


ericrsmith35:
1 - updated on unofficialsf.com
2 - same
3 - I'll try and get an example up on the website

PR #239, opened 2020-01-05
Collection global fixes

alexed1:


alexed1:
collection processor actions need to be global and not public to show up in flow builder
added PostRichChatter

PR #247, opened 2020-01-12
Quick choice dual columns

ericrsmith35:
Added an option to display the visual textboxes in a single or dual columns and cleaned up the html code

### DIFF
--- a/flow_screen_components/quickChoiceFSCproject/force-app/main/default/lwc/quickChoiceFSC/quickChoiceFSC.js-meta.xml
+++ b/flow_screen_components/quickChoiceFSCproject/force-app/main/default/lwc/quickChoiceFSC/quickChoiceFSC.js-meta.xml
@@ -8,21 +8,22 @@
     </targets>
     <targetConfigs>
         <targetConfig targets="lightning__FlowScreen">
-            <property name="value" label="Value" type="String"></property>
-            <property name="choiceValues" label="Choice Values (String Collection)" type="String[]"></property>
-            <property name="choiceLabels" label="Choice Labels (String Collection)" type="String[]"></property>
-            <property name="choiceIcons" label="Choice Icons (String Collection)" type="String[]"></property>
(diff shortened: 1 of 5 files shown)


flow_screen_components/quickChoiceFSCproject/force-app/main/default/lwc/quickChoiceFSC/quickChoiceFSC.js-meta.xml
```diff
@@ -8,21 +8,22 @@
...
-            <property name="displayMode" label="Display Mode" type="String" ></property>
-            <property name="recordTypeId" label="Record Type Id" type="String"  ></property>      
-            <property name="objectAndFieldName" label="Object Name and Field Name (dot notation)" type="String" ></property>
-            <property name="objectName" label="Object Name" type="String" ></property>
-            <property name="fieldName" label="Field Name" type="String" ></property>
-            <property name="allowNoneToBeChosen" label="Allow 'None' to be Chosen" type="Boolean"></property>   
+            <property name="value" label="Value" type="String"/>
+            <property name="choiceValues" label="Choice Values (String Collection)" type="String[]"/>
+            <property name="choiceLabels" label="Choice Labels (String Collection)" type="String[]"/>
+            <property name="choiceIcons" label="Choice Icons (String Collection)" type="String[]"/>
```
alexed1: Let's prepend "Visual Mode - " in the labels of properties that are specific to your mode for clarity. I'll do it.

alexed1:
I merged this but then ran into a cryptic error that wouldn't let me push the resulting code to a scratch org. So I reverted the merge, which means master does not have these changes.

Pull the latest master and confirm that the combined code pushes to a new scratch org without problem. Then create a new pull request and I'll try again. 

Let's rename your mode from "Visual" to "Card" for clarity. 

Note that I've removed the field ObjectAndFilename. It was duplicative  because we can solve the problem with separate object and filename attributes

ericrsmith35:
Did you run into an issue with the managed package or with the source code?

PR #257, opened 2020-01-22
SaveRecordsAsync

davidentremont:
Proposing an action I've been testing to upsert records asynchronously from flow.  Sends a collection (generic) into the invocable and returns an async job id back to the flow.

Looking for feedback since this is a new pattern for us.


alexed1:
Would you like to write a post about this action, either on unofficialsf.com or somewhere else?

PR #260, opened 2020-01-25
Record detail enhancements v1.3

ericrsmith35:
Added on option to display and handle a Cancel button when in Edit mode
Added an option to support **flow navigation** on record save and cancel                           
Included descriptive labels and help text for Flow parameters
Added another test to the flow for record creation


flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.html
```diff
@@ -1,8 +1,30 @@
+<!-- 
+Lightning Web Component for Flow Screens:       recordDetailFlowScreenComponent
 
+The RecordDetail component provides new ways to display record fields in Flow screens.
```
alexed1: Let's keep all the docs in one place. Currently that's https://unofficialsf.com/the-recorddetail-component/

I like keeping the docs on the site and not here because the code can be intimidating. 

The argument against putting it in both places is that it falls out of sync and there's more to update...

alexed1: if you're worried about prematurely publishing changes, consider create a Page (not a Post) up on the site. that way the new content can be reviewed privately.

flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.html
```diff
@@ -52,12 +74,22 @@ <h2 key={error}> {error} </h2>
...
-                                                class="slds-m-top_small slds-float--right"
-                                                variant="neutral"
+                                                class="slds-m-top_small slds-p-left_small slds-float--right"
+                                                variant="brand"
                                                 type="submit"
                                                 name="save"
                                                 label="Save">
                                         </lightning-button>
+                                        <template if:true={flowDisplayCancelButton}>
+                                            <lightning-button
```
alexed1: In the docs, can you provide a sample image and explain how the button interactions work.

flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.js-meta.xml
```diff
@@ -10,15 +10,19 @@
...
+            <property name="objectApiName" type="String" role="inputOnly" label="Object API Name" description="If you pass in no recordId, an Object API Name, and mode of Edit, it will open in edit mode and create a new record on save."/>
+            <property name="fields" type="String" role="inputOnly" label="Record Field List" description="A comma-separated list of field names from the recordId’s object type. If this is not provided, this component will display all fields accessible to the running user’s profile"/>
+            <property name="mode" type="String" role="inputOnly" label="Mode (View or Edit)"/>
+            <property name="flowNavigationOnSave" type="Boolean" role="inputOnly" label="Flow - Navigation on Save?" description="Check if you would like the flow to navigate to the Next screen (or Finish) when clicking Save.  For best results, hide the page footer in the flow screen."/>
+            <property name="flowDisplayCancelButton" type="Boolean" role="inputOnly" label="Flow - Display a Cancel Button?" description="Check if you want to display a Cancel button when in Edit Mode."/>
```
alexed1: can you think of a use case where someone wouldn't want to provide Cancel? If not, let's remove the property and just have it always on, in the interest of keeping the interface more streamlined.

flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.js-meta.xml
```diff
@@ -10,15 +10,19 @@
...
+            <property name="recordTypeId" type="String" role="inputOnly" label="Record Type Id" description="When provided, picklist values will be filter for the selected record type."/>
+            <property name="objectApiName" type="String" role="inputOnly" label="Object API Name" description="If you pass in no recordId, an Object API Name, and mode of Edit, it will open in edit mode and create a new record on save."/>
+            <property name="fields" type="String" role="inputOnly" label="Record Field List" description="A comma-separated list of field names from the recordId’s object type. If this is not provided, this component will display all fields accessible to the running user’s profile"/>
+            <property name="mode" type="String" role="inputOnly" label="Mode (View or Edit)"/>
+            <property name="flowNavigationOnSave" type="Boolean" role="inputOnly" label="Flow - Navigation on Save?" description="Check if you would like the flow to navigate to the Next screen (or Finish) when clicking Save.  For best results, hide the page footer in the flow screen."/>
```
alexed1: I like to use "Transition" to describe automatic transitions and "Navigation" to describe situations where there's some choice available. This feels a little more like a TransitionOnSave

flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.js-meta.xml
```diff
@@ -10,15 +10,19 @@
...
+            <property name="flowDisplayCancelButton" type="Boolean" role="inputOnly" label="Flow - Display a Cancel Button?" description="Check if you want to display a Cancel button when in Edit Mode."/>
+            <property name="flowNavigationOnCancelDirection" type="String" role="inputOnly" label="Flow - Navigation Direction on Cancel (Previous or Next)" description="Specify Previous or Next for the flow navigation direction after the Cancel button is clicked."/>
+            <property name="isCancelButton" type="Boolean" role="outputOnly" label="Cancel Button Selected" description="This value will be set to True is the user selected the Cancel button."/>
```
alexed1: Sometimes it's wise to allow for additional future states. If you made this text and called it requestedAction, you could set it to Cancel but could also set it to other things in the future. I"m not sure that there are other useful cases, but I generally use boolean types only when I'm really certain.

flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.html
ericrsmith35: Will do - I hadn't started on the documentation updates yet.

flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.js-meta.xml
ericrsmith35: I would think it should always be there.  I'll make the change.

flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.js-meta.xml
ericrsmith35: The option to display the Cancel button has been removed.  It will always be displayed in Edit mode.

flow_screen_components/recordDetailFSC/force-app/main/default/lwc/recordDetailFlowScreenComponent/recordDetailFlowScreenComponent.js-meta.xml
ericrsmith35: Would you leave flowNavigationOnCancel as is since there will be a selection made for which direction?  If so, would it be more confusing to rename the Save to Transition while the Cancel is Navigation?

PR #262, opened 2020-01-26
commit

professorFlow:
Alex, testing my first pull request!


ericrsmith35:
Alex - I'd like to test this first.  So far it looks like there might be an unexpected issue with existing flows that would not have an icon specified.  The display would change from no icon to a default icon unless the parameters were reconfigured for each existing flow.

ericrsmith35:
I have tested this by installing into an org that has the cuurent version of the lookup component and the lookup is not working.  No values are displayed when clicking into the input box and the console shows a stop at performLookupOnFocus.

alexed1:
Chris, congratulations on your first pull request! The kind of problem that Eric has pointed out is quite common and the whole point of pull requests is to make it easy to discover this stuff in isolation, so this is all good. At this point, you should try to reproduce the problem Eric is seeing in your local environment and scratch/test orgs. If necessary, I'm sure Eric will help you reproduce the problem.

professorFlow:
I'm looking into it now. I was able to reproduce this issue. 
If I install from the repository and then update the components it works. However, when I install the V2.3 package and then update the components it does not work.

PR #275, opened 2020-02-03
Quickchoice125

ericrsmith35:
Alex - I updated the source that was in the package hosting org.  It may have reintroduced the record.field code?

### DIFF
--- a/flow_screen_components/quickChoiceFSCproject/force-app/main/default/lwc/quickChoiceFSC/quickChoiceFSC.js
+++ b/flow_screen_components/quickChoiceFSCproject/force-app/main/default/lwc/quickChoiceFSC/quickChoiceFSC.js
@@ -1,6 +1,7 @@
 import { LightningElement, api, track, wire } from "lwc";
 import { FlowAttributeChangeEvent } from "lightning/flowSupport";
 import { getPicklistValues } from "lightning/uiObjectInfoApi";
+import TYPE_FIELD from "@salesforce/schema/Account.Type";
 
 /* eslint-disable no-alert */
 /* eslint-disable no-console */
@@ -31,6 +32,7 @@ export default class SmartChoiceFSC extends LightningElement {
 
 
 	masterRecordTypeId = "012000000000000AAA"; //if a recordTypeId is not provided, use this one
+	@api objectAndFieldName;
 	@api inputMode;
 	@api required;
 	picklistOptionsStorage;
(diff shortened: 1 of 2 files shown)


clifford-fra:
Hi Eric,

I just noticed a small "bug" in the code. The component throws a null pointer exception, if a default value is passed as parameter, which is not in the String label collection. So using:

```
//show default visual card as selected
renderedCallback() {
	if (this.showVisual && this.value != null) {
		if (this.template.querySelector('[data-id="' + this.value + '"]') != null) {
			this.template.querySelector('[data-id="' + this.value + '"]').checked = true;
		}
	}
}
```

will fix this behavior and make it consistent with how default values are handled in Picklists for example.

flow_screen_components/quickChoiceFSCproject/force-app/main/default/lwc/quickChoiceFSC/quickChoiceFSC.js
```diff
@@ -51,7 +53,8 @@ export default class SmartChoiceFSC extends LightningElement {
 	//possibility master record type only works if there aren't other record types?
 	@wire(getPicklistValues, {
 		recordTypeId: "$recordTypeId",
-		fieldApiName: "$calculatedObjectAndFieldName"
+		fieldApiName: "$objectAndFieldName"
```
alexed1: It looks like this code is flipflopping. I originally wanted to do separate object and field attributes because they're simpler to explain, but in the very first version I hadn't figured out how to use the getters and setters to properly generate a calculated version that could be used (the calculated version is then used by the getPicklistValues wire service. Once I figured that out, I removed the "ObjectAndFieldName" attribute and don't want to re-add it, as is proposed in this commit.

I suggest that you cancel this commit, pull from the current codeline (at the moment the latest commit is Feb1 https://github.com/alexed1/LightningFlowComponents/commit/efb3a491e5dbb97862202b9da80a0c05b610b39c), add your rendered callback fix, take the fix that @clifford-fra suggests, if you agree with it, and create a new pull request with just those two fixes.

ericrsmith35: Alex - I have the updates made locally but I'm not able to get them to the managed package dev org because I can delete the older versions of the test flow.  Do you have a way of removing the older flow version?

alexed1: I looked into it a bit and couldn't find a way. deleting stuff from managed packages is deliberately hard to do. all you can really do is deprecate stuff and ask the user to manually delete it. So I think we should just abandon that hosting org and create a new one. why don't we divide and conquer: go put the latest version here: 

hostquick@deved.org | test1234

I'm on the fence about whether to go managed or unmanaged. This component is getting a fair amount of updates so maybe we leave it unmanaged for now. But if we decide to, we can give hostquick a namespace and generate a managed package

ericrsmith35: Please email me the Verification Code for logging in to the new org

PR #277, opened 2020-02-06
Calculate Business Hours V1

professorFlow:
A new action for Flow Builder that allows a user to enter a start and end date, and then provide the business hours id. The action will calculate the milliseconds, seconds or minutes between the two dates in business hours.

### DIFF
--- a/flow_action_components/CalculateBusinessHours/CalculateBusinessHours.cls
+++ b/flow_action_components/CalculateBusinessHours/CalculateBusinessHours.cls
@@ -1,0 +1,42 @@
+public class CalculateBusinessHours {
+    public class inputTime
+    {
+        @InvocableVariable(label='Business Hours Id' required = true)
+        public String businessHoursId;
+        @InvocableVariable(label='Start Date' required = true)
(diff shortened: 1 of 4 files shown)


professorFlow:
@alexed1 @ericrsmith35 Please review :)

alexed1:
I also encourage you to create an sfdx project in this directory and point the readme at your blog post about this action.

cchrisv:
1. I'll start doing that!
2. I'll update the name. I have two more classes that do different operations based on business hours.  I'll make sure the names make it clear what their purpose is.
3. I'm brand new to APEX (this is my first go at it). What do you mean by this? Do you mean renaming inputTime to request and outputTime to response?
4. I'll work on this. Would this work for screen components as well?
5. I'll work on this too.

alexed1:
3. yeah, basically i'm just suggesting a different naming pattern. if you look at the other actions, they almost all use the same naming. Having two inner classes like that inside each invocable action is a little complicated for newcomers, as you may have noticed, so consistent naming might increase comfort. But not a big deal.

PR #278, opened 2020-02-06
Quick choice unmanaged 126

ericrsmith35:
Fix Visual Card default selection display v1.26


ericrsmith35:
unofficialsf.com has been updated with new unmanaged package links and the style width parameter has been added to the instructions in the post

PR #281, opened 2020-02-08
Added Multi-Lingual, test cases and others identified in pull request

### LINKED ISSUE #270: Send HTML Email: Add automatic history recording
Implement the solution below. add a boolean attribute to let user specify whether they want to log the email, and make sure it handles cases where no recordId is provided.

Megan Moody says:(Edit)
January 10, 2020 at 9:08 am
What I ended up doing was adding the following to the code below setting of the subject:

// Log the email to the feed of the record
mail.setWhatId(recordId);

Because of the business requirements, I needed to log the actual email sent to the user. The above did that. Thank you!

jdpond:
Replaces previous Pull request - cleaned.

I think this is ready to look at. Sorry it took so long - the day job got in the way. Some areas for checking:

1. **Added functionality for EmailTemplates**. EmailTemplates with the same name can have tags in their description (ex: "Language=es_MX") that distinguishes them as being a translation of all other EmailTempalates of the same name with the specified language.
2. **Modified SendHTMLEmail to use** "without sharing". This is for Community Guest (and Registered) users. There is an existing support request about accessibility (See Success Community Log) that has been running for a while. That sharing setting will be required for both Community and eventually EnhancedLetterheads when they get it fixed.
3. **Significant modifications for tests**. Code coverage now SendHTMLEmail(91% 164/179 with created records and 80% 144/179 without. Added several test and a switch TEST_WITH_CREATED_RECORDS that modifies testing as documented in SendHTMLEmailTest.
4. **Changed several handling of InvocableActionException.**  There were several places where a generated exception would not trigger a flow exception. Personally, I learned a lot from this as to which objects are valid and what can be done with them. Without this, no flow exception is thrown and the email disappears to never-never land requiring log-level debugging.
5. **Changed the RecordId parameter.** It's now 'Associated RecordId(WhatId/recordId for template and activity)'
6. **Fixed #270.** This was part of the whole mixed use of RecordId (for merge fields, templates, activities, attachments, . . .)

Take a look and tell me what you think. My email is jack.pond@psitex.com and I'd be glad to create a draft of the revised wordpress documention for approval too.

### DIFF
--- a/flow_action_components/SendHTMLEmail/force-app/main/default/classes/SendHTMLEmail.cls
+++ b/flow_action_components/SendHTMLEmail/force-app/main/default/classes/SendHTMLEmail.cls
@@ -1,16 +1,32 @@
-public with sharing class SendHTMLEmail {
+/**
+ * @File Name          : SendHTMLEmail.cls
+ * @Description        : Uses Spring/Summer '20 EmailTemplate Object + ContentVersion with multi-lingual
+ * @Author             : various, including Salesforce, Alex Edelstein, https://digitalflask.com/blog/send-email-attachments-salesforce-apex/
+ * @Group              : unofficialSF
+ * @Last Modified By   : jack.pond@psitex.com
+ * @Last Modified On   : 2/5/2020, 4:02:33 PM
+ * @Modification Log   : 
+ * @License            : LICENSE found in https://github.com/alexed1/LightningFlowComponents
+ * Ver       Date            Author      		    Modification
+ * 1.32    2/5/2020   jack.pond@psitex.com     Modified for multi-lingual and to throw InvocableActionException on exceptions
+**/
+
+public without sharing class SendHTMLEmail {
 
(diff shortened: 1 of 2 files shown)


jdpond:
Will do 2,3, and 4, but day job will get in the way the next couple of days.  THANKS!!! We really needed this Action Component!

jdpond:
About to do 1,2,4.  I noticed you added bulkification (I should have done that, sorry, just didn't want to change too much at once) and created the package and deployed it.  So 3 is done.  Working on 1,2,4

jdpond:
1,2,3,4 Done

alexed1:
minor items: the only place we should link to github pages is the View Source line. All other links should stay in the friendly unofficialsf zone, so admins aren't forced to deal with github. So can you change that first link in https://unofficialsf.com/send-html-email-1-32-now-supports-multi-language-for-lightning-email-templates/

PR #297, opened 2020-03-07
Update LightningLookupHelper.js

ericrsmith35:
Remove comment @519
This is needed so the drop-down doesn't remain displayed on selection


ericrsmith35:
Withdrawn - Not needed

PR #324, opened 2020-04-14
Datatable lwc fsc v1.1

ericrsmith35:
Added the ability to filter columns and cleaned up some error handling.

### DIFF
--- a/flow_screen_components/datatableLwcFsc/force-app/main/default/lwc/datatableLwcFsc/datatableLwcFsc.js-meta.xml
+++ b/flow_screen_components/datatableLwcFsc/force-app/main/default/lwc/datatableLwcFsc/datatableLwcFsc.js-meta.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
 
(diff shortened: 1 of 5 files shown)


flow_screen_components/datatableLwcFsc/force-app/main/default/lwc/datatableLwcFsc/datatableLwcFsc.js-meta.xml
```diff
@@ -35,11 +35,15 @@
                 description="'All' or a Comma separated list of ColID:True or False  
                   --  NOTE: Some data types cannot be edited in a datable (lookup, picklist, location, encrypted, rich text, long text area)
                   --  NOTE: ColIDs can be either the column number or the field API Name"/>
+            <property name="columnFilters" label="- Column Filters (Col#:true,...) or All" type="String" role="inputOnly"
+                description="'All' or a Comma separated list of ColID:True or False
```
alexed1, line 39: The flexibility here is respectable. I probably would have just gone with the api name because we generally discourage direct use of id's in something like this. If a flow using this is deployed to another org, it will likely still work if names are used, but will fail if id's are used. Is there a use case where you really want to use ID's?

ericrsmith35:
ColID is not a Salesforce ID.  It only refers to how the user is
identifying the datatable column.  I support both the column number
(1,2,3,...) and the field API name.

Eric

PR #332, opened 2020-04-19
Quick choice fsc v1.4

ericrsmith35:
Added help text for all configuration attributes to make this component easier to use

Fixed a bug where only the last selected visual card would display a check when multiple Quick Choice components were on the same screen


alexed1:
nice improvements

PR #334, opened 2020-04-22
Bulkify ExecuteSOQL and GetLayoutFields actions

RealEvanPonter:
When this component is executed in multiple flow interviews during a single transaction, it should iterate through the requestList instead of only acting on the first item. This solves the "The number of results does not match the number of interviews that were executed in a single bulk execution request." error

### DIFF
--- a/flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes/ExecuteSOQL.cls
+++ b/flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes/ExecuteSOQL.cls
@@ -4,12 +4,15 @@ public with sharing class ExecuteSOQL {
 
         Results results = new Results();
         List<Results> responseWrapper = new List<Results>();
-        String soqlQuery = requestList[0].soqlQuery;
-        soqlQuery = replaceWithFormattedValues(soqlQuery);
-        results.sObjects = Database.query(soqlQuery);
+        
+        for (Requests req : requestList) {
+            String soqlQuery = req.soqlQuery;
+            soqlQuery = replaceWithFormattedValues(soqlQuery);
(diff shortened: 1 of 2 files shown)


RealEvanPonter:
@alexed1 Awesome - yeah I can take a look at some of the others. I'm actually working on a deployment for a client and was hoping to install the latest version of the ExecuteSOQL package (which is an unlocked package and doesn't seem to allow me to move my updated code via change set into other environments). What is the process for creating a new version of this package? Let me know, thanks

RealEvanPonter:
Hi @alexed1 Just wanted to follow up and see if you could help me upgrade the Execute SOQL package to the latest version - I am trying to deploy this bug fix to a Production environment for a client so that their flows will process multiple records in a single transaction. Please let me know what needs to be done in order to create a new version of the unlocked package, or if you have any tips for deploying an edited apex class that is part of an unlocked package.

alexed1:
Hi, Evan. Version 1.2 is bulkified and can process multiple records. https://unofficialsf.com/execute-soql-v1-2-handles-flow-dates-properly-and-works-with-flow-schedule-and-record-change-triggers/

Did you try to deploy that version? 

I'm not an expert in change sets, but my general understanding is that you don't typically mix packages and changes sets. You don't use the change set process to deploy a package.  It's either one or the other. Take a look at https://trailhead.salesforce.com/content/learn/modules/application-lifecycle-and-development-models

If this doesn't help, I suggest you post to the Trailblazer Community which has lots of experts in change sets. Ironically, we product managers don't work in an environment similar to yours, and we only use the latest packaging approaches, so we have some blind spots.

RealEvanPonter:
Thanks @alexed1 - I've got version 1.2 installed and working now

PR #347, opened 2020-04-29
MODIFY: Fix Regex Search in ExecuteSOQL to Handle Nested Queries

sillyrags:
Primary Fix:
- Update regex for SOQL query string to to handled nested relationship queries after the WHERE clause; The code will first check to see if one of those queries exists and if true, will recursively search the nested queries and do the same formatted on those; If false, it will continue on with the single query search

Secondary Fixes:
- Strip unwanted formatting characters from the query string that come in when text template is used in flow: '\n', '\r\n', '\t'
- Fix time stamp formatting on date strings with a time: '00:00 am' is an invalid format. The equivalent, valid format using regular time is '12:00 am'. '00:00' is only valid for 24hr formatting.
- Fix month name to month number conversion: prepend a 0 on the month number if the string value is < 10

Improvements:
- throw some exceptions for more informative error handling
- cleaned up unit suite and fixed failing tests

Potential issues not yet fixed:
- It appears that after the ' order by ' substring is removed from the query, it is never added back before querying.

### DIFF
--- a/flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes/ExecuteSOQL.cls
+++ b/flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes/ExecuteSOQL.cls
@@ -1,26 +1,19 @@
 public with sharing class ExecuteSOQL {
+    class ExecuteSOQLException extends Exception {}
     @InvocableMethod
     public static List <Results> getEligibleProducts(List<Requests> requestList) {
-
         Results results = new Results();
         List<Results> responseWrapper = new List<Results>();
-
         for(Requests curRequest : requestList) {
-
             String soqlQuery = curRequest.soqlQuery;
             soqlQuery = replaceWithFormattedValues(soqlQuery);
             results.sObjects = Database.query(soqlQuery);
-    
-    
             responseWrapper.add(results);
-            
         }
         return responseWrapper;
-       
     }
-
     public static String replaceWithFormattedValues(String soqlQuery) {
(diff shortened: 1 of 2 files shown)


flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes/ExecuteSOQL.cls
```diff
@@ -29,53 +22,66 @@ public with sharing class ExecuteSOQL {
...
             while (mMatcher.find()) {
-                String whereClause = mMatcher.group(0);
-                fieldNames.addAll(whereClause.split('\\(|\\)|>=|<=|!=|=|>|<| in\\(| like | in:| or| and'));
+                String whereClause = mMatcher.group(0).replaceAll('\\(select .+\\)', '(SUBQUERY)');
+                fieldNames.addAll(whereClause.split('\\(|\\)|>=|<=|!=|=|>|<| in\\(| not in\\(| like | in:| or| and'));
             }
-
             if (!fieldNames.isEmpty()) {
-                for (Integer i = fieldNames.size() - 1; i >= 0; i -= 2) {
+               for (Integer i = fieldNames.size() - 1; i >= 1; i -= 2) {
```
sillyrags, line 58: For good measure. If an unhandled query string makes it to here, you would get an out of bounds error if you let `i` go to 0.

sillyrags:
@alexed1 I added a couple of example queries to the PR description.

PR #348, opened 2020-04-30
Revert "MODIFY: Fix Regex Search in ExecuteSOQL to Handle Nested Queries"

alexed1:
Reverts alexed1/LightningFlowComponents#347

Ragan, getting this error when I try to package the merged code:
 ExecuteSOQLTest: Dependent class is invalid and needs recompilation:
 Class ExecuteSOQL : Variable does not exist: cleanObjectName,ExecuteSOQL: Variable does not exist: cleanObjectName

looks like cleanObjectName is missing a definition.


sillyrags:
@alexed1 Yes, I missed that when moving the code over here for the PR. Here is the PR to fix the issue: https://github.com/alexed1/LightningFlowComponents/pull/349

PR #352, opened 2020-05-04
MODIFY: Add support for date literals in query parameters (ExecuteSOQL)

sillyrags:
ExecuteSOQL action fails when date literals are provided as query parameters for date fields. This PR adds support for date literals and adds associated unit test for validation.


flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes/ExecuteSOQL.cls
```diff
@@ -7,14 +8,15 @@ public with sharing class ExecuteSOQL {
...
+            System.debug('Logs: query: ' + soqlQuery);
             results.sObjects = Database.query(soqlQuery);
             responseWrapper.add(results);
         }
         return responseWrapper;
     }
     public static String replaceWithFormattedValues(String soqlQuery) {
         String endingClause = '';
-        soqlQuery = soqlQuery.toLowerCase().replaceAll('\r\n|\n|\r|\t',' ');
+        soqlQuery = soqlQuery.replaceAll('\r\n|\n|\r|\t',' ');
```
sillyrags, line 19: Bonus fix: by lower casing the query string rather than doing a case insensitive regex search, queries like this break because Id is a case-sensitive field: 
```
Select Id from Contact where AccountId = 'x'
```

alexed1:
can you provide an example of a couple of queries that didn't work before this fix, but work now? I'll put them in the docs.

sillyrags:
Query on Id string literal:
`SELECT LastName from Contact WHERE Id = '1123xxx333e3e'`

Query on date literal:
`SELECT Id from Contact WHERE CreatedDate = LAST_n_DAYS:14`

sillyrags:
@alexed1  I updated the comment field in the PR with the two examples.

PR #358, opened 2020-05-07
New Input variable to store api names of fields to be excluded from page layout

aleksbasic2:
Feature to exclude fields that are in the page layout from appearing in the flow component.

Classic use case is for standard fields that cannot be removed from page layout (like Case Origin or Contact Name).

Account object and Account Layout were failing the test class on my org (not sure why exactly), but it passed when I changed it to Case.

### DIFF
--- a/flow_action_components/GetLayoutFields/force-app/main/default/classes/GetLayoutFields.cls
+++ b/flow_action_components/GetLayoutFields/force-app/main/default/classes/GetLayoutFields.cls
@@ -11,6 +11,7 @@ public with sharing class GetLayoutFields {
         for (Requests req : requestList) {
             String layoutName = req.layoutName;
             String objectName = req.objectName;
+            String excludeFields = requestList[0].excludeFields;
             String layoutFieldsCSV = '';
 
             layoutName = objectName + '-' + layoutName;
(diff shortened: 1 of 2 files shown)


alexed1:
I'm posting about your contribution. Want to give me any more info, such as your full name and/or a website to point people at?

alexed1:
Note this error: 
GetLayoutFields: Defining type for global fields must be declared as global,GetLayoutFieldsTest: Dependent class is invalid and needs recompilation:
 Class GetLayoutFields : Defining type for global fields must be declared as global,GetLayoutFields: Enclosing type for global fields in apex classes must be declared as global

It's an easy fix, and I'll do it right now.

aleksbasic2:
Thanks! Noted.

It would be cool if you mention my name :)

-Aleksandar Basic

solo-1234:
Was the 'global fields' thing fixed yet? I'm running into the same issue - I tried updating my GetLayoutFields and test classes to the newest version but it doesn't let me save them.

Thanks!

aleksandarbasic:
https://github.com/alexed1/LightningFlowComponents/pull/368
created a pull request to update global to public.

Looks like my classes had global access modifiers for all variables and methods instead of public as the source code does.

solo-1234:
Thanks!

PR #360, opened 2020-05-13
Support for Boolean, Currency, Date, and Datetime to SortCollection

hayeswise:
I've added support for Boolean, Currency, Date and Datetime to SortCollection.  I performed simple tests on the four data types and using the Date sort in a new flow I'm developing.


flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
```diff
@@ -21,7 +21,7 @@ global with sharing class SortCollection {
             }
```
alexed1, line 21: This is great! Welcome to the project. Very glad to have your contribution

flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
```diff
@@ -21,7 +21,7 @@ global with sharing class SortCollection {
...
 
             if (!outputCollection.isEmpty()) {
-                Map<String, String> fieldTypes = CollectionProcessorsUtils.getFieldTypes(outputCollection[0], allFields);
+                Map<String, String> fieldTypes = getFieldTypes(outputCollection[0], allFields);
```
alexed1, line 24: Here, this action was using a single shared getFieldTypes, which provides some reuse. What was the reason you decided to create a copy of the code here?

flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
```diff
@@ -97,13 +97,89 @@ global with sharing class SortCollection {
...
+                    if (origValue && !compareToValue) {
+                        returnValue = 1 * direction;
+                    } else if (!origValue && compareToValue) {
+                        returnValue = -1 * direction;
+                    }
+                }
+             } else if (fieldType == 'CURRENCY') {
+                Decimal origValue = Decimal.valueOf(acc.get(this.field)+'');
+                Decimal compareToValue = Decimal.valueOf(compareToAcc.acc.get(this.field)+'');
+                if (origValue == null && compareToValue == null) {
```
alexed1, line 119: It looks like there's an opportunity to DRY this if statement up, as it appears to be repeated without variation. Could you put it into a separate function and have all fo the field type sections call that function?

flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
hayeswise: The Sort has been a very helpful component to a flow that I'm building. Thank you!

flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
hayeswise: I don't recall working on that section of code - I'm thinking it was from an older version of SortCollection.

flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
hayeswise: Yes. I'll see what I can do!

flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
alexed1: let's stick with the version that's in ColectionProcessorsUtils in that case.

flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
hayeswise: I've added support for other field types and have been running Apex unit tests. I need some help though: When creating custom fields, which field types corresponding to DECIMAL and INTEGER?
E.g., NUMBER(10,8) corresponds to DOUBLE.
Thanks!

ericrsmith35:
The following Apex types are all number - 'DECIMAL', 'DOUBLE', 'INTEGER',
'LONG'

hayeswise:
@ericrsmith35 I wasn't clear enough in my question - by "custom fields", I meant fields added via the Object Manager.  I later stumbled across this reference:
[API Data Types and Salesforce Field Types](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/api_ui_datatype_map.htm)
https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/api_ui_datatype_map.htm

flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
hayeswise: The code has been updated and tested. Do I submit another pull request?
Testing via SortCollectionTest.cls and spot checking using the FlowBuilder.

hayeswise:
Thanks.  I enjoyed the project!  I also think it would be useful to update the documentation.  However, the number of field types support has grown considerably. And I also worry about readers trying to map the API data types to the UI field types - which might be useful to document in the GitHub wiki for this project.

As documented, I was unable to test for decimal, integer, and long.

Here's what I would like to see for the documentation, although it may be a bit verbose. Please feel free to edit to meet your sensibilities.

##Sort Collection
Takes a collection and sorts it per the provided sort keys. Null/empty SObject field values are considered smaller than non-null/non-empty values.  Sort Collection supports 16 API data types and 20 UI field types.

Notes:

1. Picklist fields are sorted by API name, not in picklist order, and without concern to picklist value active/inactive status.
2. The sorting priority may be reversed from what you are accustomed to. It sorts first on the last key pair and then works backwards to the first.

###Inputs
List<SObject> inputCollection;
A collection of SObjects.  Be sure that the objects include any field referenced in the Sort Keys.

String sortKeys;
One or more comma separated sort keys formatted as [Field API Name]:[ASC|DESC]. For example, CloseDate:ASC,Amount:DESC.
Sort directions other than DESC are treated as ASC. Invalid sort key field api names will result in a "System.NullPointerException: Attempt to de-reference a null object". If the field is missing from an SObject in the collection, the following error will occur: "System.SObjectException: SObject row was retrieved via SOQL without querying the requested field: [Object Name].[Field Name]".

###Outputs
List<SObject> outputCollection;

PR #361, opened 2020-05-15
ADD: New Dedupe Records Actions

sillyrags:
Adds two new dedupe records actions--one that returns a list of unique field values (null reduced) given a list of records and field name to dedupe on, and one that returns an updated list of records given a list of records and field name to dedupe on.

Errors thrown when: 
(1) you provide a null or empty list of records as input
(2) you provide a null or empty field name
(3) the field name provided is invalid

Adds supporting unit tests with 95% code coverage.


sillyrags:
@alexed1 This action extends the functionality of `ExtractFieldsFromCollection` by adding the ability to dedupe the list of values, and supporting bulkification. Might be worth considering merging these two at some point. 

https://github.com/alexed1/LightningFlowComponents/blob/master/flow_action_components/CollectionProcessors/force-app/main/default/classes/ExtractStringsFromCollection.cls

flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes/ExecuteSOQL.cls
```diff
@@ -1,4 +1,5 @@
-public with sharing class ExecuteSOQL {
```
alexed1, line 1: is it intentional that there are changes to ExecuteSOQL in this pull request? My guess is not....

flow_action_components/DedupeRecordCollection/force-app/main/default/classes/DedupeRecordCollection.cls
```diff
@@ -0,0 +1,48 @@
+global with sharing class DedupeRecordCollection {
+    global class DedupeRecordException extends Exception {}
+
+    @InvocableMethod(label='Dedupe Records from Collection' description='Dedupe a list of records using a given field as the unique value and return the unique collection of records' category='Util')
```
alexed1, line 4: The two actions are pretty similar so we should give thought to how someone coming at this for the first time would figure out which to use. There's an argument to be made for combining them into one action which would return both kinds of output. But that does require a user who just wants a simple solution to specify an output collection type even if they're not going to use it, which may be confusing.
I think I'd name the actions to focus on their differentiation:
GetUniqueFieldsFromCollection
DedupeRecordCollection

flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes/ExecuteSOQL.cls
sillyrags: Good catch.. There shouldn't be any changes to SOQL in the PR. This is the version I am using on my local. :)

sillyrags: I updated this branch from the remote and discarded my local changes, so these changes are gone now.

flow_action_components/DedupeRecordCollection/force-app/main/default/classes/DedupeRecordCollection.cls
sillyrags: I agree @alexed1 , and per my initial comment on the PR, `DedupeFieldsFromCollection` (your proposal as `GetUniqueFieldsFromCollection`) extends functionality of the existing `ExtractFieldsFromCollection` by providing that `dedupeValues` parameter, plus a few extra enhancements. My inclination would be to merge those two and then add the `DedupeRecordCollection` as a new action. 

FWIW--we did start this as a single action, but figured the uses cases for each were mutually exclusive.

alexed1: Agree that at this point the uniqueness chops should probably just be added to ExtractFieldsFromCollection. Please feel free to make that change if you have time.

alexed1:
If you have the chance, please update these pages:
https://unofficialsf.com/list-actions-for-flow/
let me know if you need editor access, Ragan.

PR #388, opened 2020-06-09
stored comma-delimited string into list var

aleksandarbasic:
instead of searching if the single string of fields separated by a comma contains each field on the layout, the inputted string of comma separated field names is stored in a string List and the list is searched.

for example, previously a user input of "Account_Status__c" would exclude the standard "Status" field since the input string contains Status. Now the contains checks if the list contains an "Account_Status__c" as one of its list members.


aleksandarbasic:
Also - fixed the wrongful declaration of excludeFields variable

PR #394, opened 2020-06-16
Sortcollection 100 percent test coverage

hayeswise:
1. Added @TestVisible annotation SortCollection's getComparable () 
2. Added Datetime test on Contract object
3. Added additional field type tests with getComparable () now test visible
4. Commented out test setup and tests using custom Sobject - no longer needed due to item 3.


hayeswise:
Closing pull request - sortCollection is not bulkified which is making it difficult to use in conjunction with a process that triggers after a data load.  I'll upgrade the code, test, and resubmit.

Issue #396, opened 2020-06-17
SortCollection Not Yet Bulkified - FLOW_ELEMENT_ERROR 


hayeswise:
Getting FLOW_ELEMENT_ERROR|The number of results does not match the number of interviews that were executed in a single bulk execution.

Error Log:
```
09:46:08.128 (1131698951)|CODE_UNIT_FINISHED|SortCollection.sort(List<SortCollection.Requests>)
09:46:08.63 (1132647321)|FLOW_VALUE_ASSIGNMENT|28853ff7d8227c67093f24cae705172c325886c-3314|Sort_Opportunities|false
09:46:08.63 (1132835605)|FLOW_VALUE_ASSIGNMENT|28863ff7d8227c67093f24cae705172c325886c-3313|Sort_Opportunities|false
09:46:08.63 (1133375697)|FLOW_ELEMENT_ERROR|The number of results does not match the number of interviews that were executed in a single bulk execution request.|FlowActionCall|Sort_Opportunities
```

Fix along with tests provided in pull request #399.

alexed1:
fixed and merged. will be part of the next package update.

PR #399, opened 2020-06-20
SortCollection Bulkification

hayeswise:
1. Bulkify SortCollection - See issue #396.
2. Testing for bulkification and additional tests to achieving 100% test coverage.


flow_action_components/CollectionProcessors/force-app/main/default/classes/SortCollection.cls
```diff
@@ -2,37 +2,35 @@
  * Provides a Salesforce Ligntning Flow component for sorting collections.
```
alexed1, line 2: This is great! It's one thing when someone submits new features. That's always nice. But it's _really_ great when someone hardens an existing extension. Thanks for the contribution!

PR #411, opened 2020-07-08
Added instantNext property suggestion, to allow Immediate Next navigation without Footer.

NickSauer7:
Especially for Visual Card, the usage may be to select and immediately advance instead of relying on Flow navigation in footer.

Added 'instantNext' inputOnly property (defaulted to false).  If set to true in Flow, then when making a selection on Visual Card, Picklist or Radio Group, the Next navigation is called.  Description makes clear this should be used only when this is only component on screen.

Example use case is a Flow that branches to sub-processes.  'Parts Manager' that branches into 'Order Parts', 'Return Parts', or 'Scrap Parts' based on the Visual Card selected.

Tested on Visual Card, Radio Button Group and Picklist against existing component.


ericrsmith35:
This is very handy in many cases but keep in mind that it does not support
any users with needs for special adaptations (sticky keys, screen reader ,
etc)

NickSauer7:
OK thanks Eric - is there a good way to allow for those considerations in such a use case?  I'm not familiar with those design elements, so let's just close/cancel this pull request, and I can circle back later after I've investigated further.

ericrsmith35:
I think it's a good pull request.  It just needs to be noted that it does
not adhere to accessibility standards.

NickSauer7:
Great, sounds good.  Where do you usually prefer to note that -- just want to follow your standards here as I'm a newbie!

ericrsmith35:
Nick,

Do you have edit access to unofficialsf.com?  The update should be added to
the instructions here:
https://unofficialsf.com/quick-choice-improved-picklists-and-radio-buttons-for-flow/
.

Eric

NickSauer7:
I only have 'View' access for existing posts (and can author new posts).  Editing someone else's posts is not an option with my current permissions.

ericrsmith35:
Send me what you would like to have updated and I'll get it in there for
you.

NickSauer7:
I was going to edit to include the following in the notes.

______________________________________________________________________________________________________________________________________
Features Section:
9.  Allow selection to prompt Navigation to Next Flow screen (Optional)

General Attributes Section (Table row addition):
Attribute | Type | Notes
instantNext | Boolean | Set to True if you want selection to invoke Flow 'Next' navigation.  Default is False.  Do not use unless component is the only one on the Flow Screen.  Note that this does not adhere to accessibility standards at this time.

_______________________________________________________________________________________________________________________________________
Just let me know if this change isn't worthy of a new release, re-packaging, etc. and can wait to include as part of larger update.   Users loved the implementation of the original, but first request was to auto-advance, so thought it may be useful.

ericrsmith35:
At this point, I think you should look at the source for the new Version
2.0 for the Summer release and see how the update could be applied against
that code.

https://unofficialsf.com/preview-quick-choice-2-0-with-a-custom-property-editor/

PR #412, opened 2020-07-11
multiMapFSC - Updated map component for LWC & Added couple toggles.

NickSauer7:
Initial version drafted here and tested for all attributes.

Updated version of map FSC Aura component.  Added address vs. geolocation toggle, list view on/off toggle, and marker selection for output from component.  

Draft post with additional details found on wp drafts:  map-with-markers-in-flow


flow_screen_components/lightningMapFSC/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Summary:
+1)	Pass in Address or Geolocation and view on Map
+2)	Adjust Title of the Map
+3)	Show or hide the List View on the right side.
+
+### Admin Instructions:
+##### Enabling and Disabling Features in Flow:
+1)  Set Geolocation attribute to true if you plan to pass in Latitude and Longitude.  Otherwise you will pass in address.
+2)  For markerList (Address), you must pass in semi-colon delimited collection of strings in the following order:  City;Country;PostalCode;State;Street;Name;Icon;MarkerValue.
```
alexed1, line 17: I like to look for ways to create inputs that simplify the user's work. In this case, some options that occur to me: If I collect data in an upstream Screen element using the Address screen component, is there a way I could use that here without having to assemble the semicolon list? Perhaps have a bunch of additional text input fields?

flow_screen_components/lightningMapFSC/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+|Parameter	               |I	 |O	     |Information 
+|--------------------------|-----|-------|-------------------------------------------------------------------------------------------------------------------------------------------|
+|**markerList**            |X    |       |List of String.  Pass in Address attributes or Latitude/Longitude depending on geolocation flag.  Default is address.  See instructions.   |
+|**title**	       		   |X	 |       |String.  Title to appear above the map                                                                                                     |
+|**showListView**	       |X	 |       |Boolean.  Set to true to show the List View on the map.  Default is false (hidden)                                                         |
+|**geolocation**	       |X    |	     |Boolean.  Set to true if you will pass in Geolocation instead of Address. Default is false                                                 |
+|**selectedMarkerValue**   |	 |X	     |String.  This is the Id of the marker selected.                                                                                            |
```
alexed1, line 30: I don't know what a marker ID is or where I'd find that, so a little more context would be great on markers

flow_screen_components/lightningMapFSC/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+### User Instructions:
+1)	Note that markers will not appear if there are address/location attributes missing.  This can cause the List View not to match the map.
+2)  If enabled by your administrator, selecting a Marker on the map or list view will capture the selection you have made.
```
alexed1, line 34: a small example describing how you might use a selection later in the flow might be good

flow_screen_components/lightningMapFSC/README.md
NickSauer7: We could do that also (I believe the current mapFSC component you have in the collection does this).  In the other Aura component from blog post by Terence he went with the collection approach, and that's what I used as the launching point for the LWC update.  The reason is that, in thinking through use cases, I see many more instances where a collection of markers would be desired as opposed to a single marker based on an earlier entry.  Perhaps there is one that I'm missing though, or misunderstanding - highly likely!  

In doing this transposition, I could see a couple future enhancements that might be useful --
1)  Search for some address value(s), find matching records of some type and then display those on the map.
2)  Add filter/search logic on the same LWC screen to pare down results.

flow_screen_components/lightningMapFSC/README.md
NickSauer7: Updated to try to clarify.

flow_screen_components/lightningMapFSC/README.md
NickSauer7: Good idea - added this.  Pushing readme update now.

NickSauer7:
mapFSC name is taken by the Aura component, and cannot push into org as LWC also, so have left as multiMapFSC.  Just let me know if missing something there.

alexed1:
in the future, push the entire sfdx project and not just the force-app subdirectory. that allows people like me to simply open the project and deploy it to a scratch org. In this case, I'll create the project around your force-app subdirectory

alexed1:
Here's a package. can you test it out and let me know if it's good to go? Can you also draft a blog post as well that shows it in use?
https://login.salesforce.com/packaging/installPackage.apexp?p0=04tf4000003RR3lAAG

alexed1:
Thanks again for your great contribution, Nick.

NickSauer7:
Thanks Alex- will do on the package in the future.  I am away from a system for the next week but will knock out the package testing upon my return.  There should be a draft blog post drafted already:  'map-with-markers-in-flow'.  Thanks again - learning a ton!

NickSauer7:
Hi Alex - I've installed the package and tested the different toggles/outputs/etc.  Made a few edits to the blog post as well.
'Present a Map with Markers in Flow'.  Just let me know of any updates you'd like to that.

This is a component that could use a lot of enhancements, and I've had a couple discussions on it so far.  Will just be a matter of time, but at least gives a launching pad for those.

PR #401, opened 2020-06-22
recordDetailsFSC : Few Enhancements

mhamzas:
- Save and Cancel Button Labels are now changeable
- Option to hide CANCEL button
- Column Size option (Show all fields in 1 -> 12 sections)
- Disable Toast Option
- Removed slds-box class to remove double boxes under flow screen
- label-stacked variant added on input fields.

PR #413, opened 2020-07-12
Add padding to layout items for readability

swood09:
Given the spacing between the field label and the actual input element, screens become difficult to use especially when there are many fields presented to the user. This is especially problematic with the two-column layout, as the input elements of fields from column 1 are adjacent to the labels of column 2.
This change simply adds padding to the layout items so that fields are a bit more distinguishable from each other. It's a small quality of life improvement, but makes a big difference.


alexed1:
This looks promising. I love small useful polish like this. Can you post before and after images here to show the impact?

swood09:
@alexed1 Of course! Screenshot attached.
I just read through @mhamzas PR #401 so would be curious to know how use of the 12 column layout will work with this.
Screen Shot 2020-07-13 at 11 57 48 AM

alexed1:
I merged in Hamza's items. can you test your fix with the merged code? Once you confirm you're still happy with it, I'll merge yours in too and build a new package.

alexed1:
Hi. Just checking in on this.

mhamzas:
Added an option to allow adding classes by your own + set the default class back to :  'slds-box slds-theme_default' 
@alexed1

swood09:
Sincere apologies that I've neglected to respond @alexed1. Work has refused to slow down for the summer!
I'm going to close this out since @mhamzas opened another PR with support for adding slds classes to the parent div

PR #416, opened 2020-07-16
Input Rich Text - Fix 2x Value Bugs, & Add Character Count, Max Characters and Labels

NickSauer7:
Labels:  Properties have labels added.
Character Count:  Default character counter
Character Limit:  Optional max characters and validation.


alexed1:
nice fixes and improvements. sorry it took so long to process this.

PR #441, opened 2020-08-08
Packaging update + turned of object create in test

jdpond:
SendHTMLEmailTest has an option to more thoroughly test by creating records.   When enabled, coverage =93% when disabled, 83%.  Creating these records may cause a failure on upload if the installing org has created custom required fields, so should not be included in the distribution.

This release turns off the record create in the test classes and adds more sophisticated package creation definition.
`    private Static Final Boolean TEST_WITH_CREATED_RECORDS = false;
`

The r[elated package installation]( https://login.salesforce.com/packaging/installPackage.apexp?p0=04t3h0000045qf3AAA) (for sendHTMLEmail).

The minimum dependent install for [Flow Based Components](https://unofficialsf-psitex-dev-ed.lightning.force.com/packagingSetupUI/ipLanding.app?apvId=04tf4000004Pu2iAAC).


jdpond:
Can't promote with the same major.minor.patch number.   Creating new pull request

PR #443, opened 2020-08-09
feat: bulkify SendHTMLEmail component

msrivastav13:
@jdpond @alexed1 This PR fixes below 

1. Bulkified DML on tasks.
2. SOQL Queries is avoided in FOR loops with dynamic SOQL and Maps.
3. Simplified logic for error handling and sending Emails. Using the Base Email class than unnecessary splitting between MassEmailMessage and SingleEmailMessage. Both of them can coexist now and work with a single Email Send invocation.
4. Improved Efficiency of the component.  One can send now 1000s of emails using this component without any issues. Of course, they will hit apex heap or Email limit if they go above 5K records.
5. Added Prettier to the code so the code formatting is consistent between different writers.
6. Added brackets around if and else statements.
7. Added a couple more asserts to the apex test code.
8. Updated Version numbers to 49.0 to the latest.
9. Removed not needed wire import from the component
10. Enhanced the response output now returns the recordId as well and moves the Task insert error to another variable.


alexed1:
Ok, here's a package for version 2.0, which incorporates both Jack's and Mohith's changes:
https://login.salesforce.com/packaging/installPackage.apexp?p0=04tf4000004Pu6BAAS
@msrivastav13 @jdpond  Can you do some quick testing of it to make sure it works properly?

msrivastav13:
@alexed1 will do some testing now with this! Thank you

jdpond:
Doesn't upgrade

jdpond:
@msrivastav13 THIS IS EXCELLENT WORK - thank you!!

msrivastav13:
Are we using unlocked package for packaging or is it 1GP packaging?

I can look into GACK when I am on computer today!

jdpond:
Fails test // @Test t014_coverageTestActivityAndBulkification: Can allow email activity and task activity on recordId object if records, plus bulkification

Examining now why.

msrivastav13:
@jdpond make sure that you have the latest from my branch for the tests! All tests successfully passed in my org with latest tests in my branch!

msrivastav13:
ok so the only diff I found when tests fail is when `    private static final Boolean TEST_WITH_CREATED_RECORDS = false;
`
 is turned on

jdpond:
**/
    private Static Final Boolean TEST_WITH_CREATED_RECORDS = true;
//
image

jdpond:
**/
    private Static Final Boolean TEST_WITH_CREATED_RECORDS = false;
//
image

jdpond:
@msrivastav13 - did you change sendHTMLEmailTest too?  It looks like several bulkification tests have been disabled.

msrivastav13:
@jdpond No, I did not disable any tests, I added more asserts and removed system. debug statements as it consumes resources. Wherever I took the system. debug i have replaced with System.asserts.

jdpond:
image

jdpond:
Let me take a deeper look - I think these may be the result of trying to test bulkification using the activity/tasks

msrivastav13:
Here is what it shows in my org

Screen Shot 2020-08-09 at 3 00 06 PM

msrivastav13:
Here is the tests in my org when `    private static final Boolean TEST_WITH_CREATED_RECORDS = false;
`

Screen Shot 2020-08-09 at 3 02 43 PM

This one shows error as you are getting! I think this may be something to do with nullification but not sure how this succeeds by changing TEST_WITH_CREATED_RECORDS to true.

I will also take a look into depth on this.

However the packaging issue is lot different than this i feel.

jdpond:
Looking into it too - I'm sure it's something minor.  Remember, the responses are bulkified too - each request has its own, unique response.

msrivastav13:
@jdpond yes looks something minor! We would just need to make sure WhatId is not userId in the Task creation. It needs to be Account, opportunity or custom object.

msrivastav13:
@alexed1 I am getting this when trying to install the package

1.  Invalid Upgrade.
The package you're installing depends on package 'FlowBaseComponents', version '1.2'. Install package 'FlowBaseComponents' in the target org before you install package 'SendHTMLEmail2'.

How can i get FlowBaseComponents package and why is SendHTMLEmail2 dependent on it?

jdpond:
https://unofficialsf.com/introducing-flowbasecomponents/
https://login.salesforce.com/packaging/installPackage.apexp?p0=04tf4000004Pu2iAAC

msrivastav13:
@jdpond Do you know why it is dependent?

jdpond:
The new version uses CPE.  CPE uses Comboflow Boxes and several other listed base components.

msrivastav13:
@jdpond Thank you! makes sense!!

msrivastav13:
@alexed1 I was able to install version 2 and works perfectly fine in my org with some smoke tests!!

jdpond:
Does not work with bulkification in org tests.   Non-Bulk (single requests) works.  Multiple requests aren't working.

jdpond:
A bit of history.  Several users were/are using a single invocation with multiple requests.  There are two types of bulkification we may have conflated here:
1.  Bulkification of multiple requests into a single invocation of the action.  Example:  [{request1},{request2},. . . }
2. Internal bulkification inside each request to optimize processing

Scenario 1 no longer works.

msrivastav13:
@jdpond Lets open issues for the ones that do not work with repro steps. I have some time next week to dig into them. 

The mixed request use case and individual requests should still function. If this is a bug I will take a crack at it.

msrivastav13:
@jdpond I am digging into tests failure for the t009 tests

` testReq.recordId = validEmailTemplate.Id;
        if (TEST_WITH_CREATED_RECORDS) {
            testReq.templateTargetObjectId = emailCreateTestContacts(
                    TEST_CONTACTS_COUNT
                )[0]
                .Id;
            testReq.recordId = emailCreateTestAccount(
                    1,
                    testReq.templateTargetObjectId
                )[0]
                .Id;
            testReq.saveAsActivity = true;
        } else {
            testReq.templateTargetObjectId = UserInfo.getUserId();
            // if templateTargetObjectId type = user, saveAsActivity must be false
            testReq.SendTOthisOneEmailAddress = EMAIL_NOSUCH;
            testReq.saveAsActivity = false;
        }`

The `recordId` when `TEST_WITH_CREATED_RECORDS = false` is passing a emailTemplateID which does not sounds like we are passing right test data and hence the test is failing.

What is `recordId` for? I thought it will be AccountId and equivalent to whatID. I think if we have clarity on this should fix it.

jdpond:
https://unofficialsf.com/send-rich-email-with-the-new-sendhtmlemail-action/comment-page-2/#comment-1278  this will blow out.

jdpond:
I am updating the documentation now - and I agree it is VERY confusing - which was one of the main reasons we delayed the release until Summer '20 and FlowBasedComponents for CPE.  Hopefully the CPE version of this makes it a lot clearer.  The weird thing is that the EmailMessage class refers to it differently (whatId,recordId) while the templates refer to it as "related" or even the type of the related object itself depending on which template type you use.

msrivastav13:
@jdpond Can you share the docs with me? sorry I am not aware. I am using both Mass Email and Single Email messages both in one go.

The returned results are not in exact order but we can make this in the same order as well if that fixes things by adding a request number that's internal to each request and then sorting the returned response order in that order.

The aim of my refactor was to optimize everything and I think code is truly optimized at this point. It's just I need to know those bugs you are discovering and should be fixable easily without altering the current design.

jdpond:
Optimization is great, but we've eliminated required functionality - some of the weirdness in the bulkification was for request bulkification - not internal code bulkification.  The design document:
[External Services – UnofficialSF.pdf](https://github.com/alexed1/LightningFlowComponents/files/5048368/External.Services.UnofficialSF.pdf)

jdpond:
Remember also that each request has its own set of parameters which are not, and should not be shared or used by other requests.  That is why you see some queries per request.

msrivastav13:
I do not agree!!! let's chat over a call sometime next week. If requests are separated they can use multiple of these components. Ideal use case one should be passing with a record lookup element with multiple records as requests.

Breaking this into different entities will also mean it limits to 10 email sends per context less than ideal. 

Also, I am looking into the CPE thing highly confusing interface was my first reaction.

The test class bugs are simple and will fix soon seems straight forward and due to confusion in what the recordId is defined!

jdpond:
You may well be right.  Might I suggest you express this to the community already using this functionality?  A great place to start would be here: https://unofficialsf.com/send-rich-email-with-the-new-sendhtmlemail-action/comment-page-2/#comment-1235

msrivastav13:
Thanks, @jdpond, let me fix some tests we have failing and make some more enhancements and then point community to do it right! With 2.0 they should be able to use for all use cases.

I need clarification on a couple of points

- What is different between saveasTask vs saveAsActivity?

- What is the recordId variable for? what should be input there by the user? 

- Also, CPE does not allow me to use bulk records? I used CPE and tried connecting the to bulk records and I not able to map.Am I missing something?

jdpond:
I think the documentation I'm working on right now will go a long way to answering 1 and 2 (above).
3) Nope, it doesn't right now bulk is only used in the manner created/proposed by [Josh ](https://unofficialsf.com/send-rich-email-with-the-new-sendhtmlemail-action/comment-page-2/#comment-1235) and is only used by folk trying to get around the functional limitations of Messaging.sendEmail().  By segmenting into batches of 10, they can use up to the full email limits.

Let's talk about whether we should add a collection of requests to the CPE - I've been mulling that one for a while and have come to no confirmed conclusion yet - you could definitely help on this design decision.

alexed1:
I'm going to revert this merge so we can align on the best way forward.

msrivastav13:
@jdpond I appreciate all of the knowledge sharing so far! Thanks for your wonderful contributions!!

The proposed solution by Josh seems to be a workaround! There is no scalability to it and one can do it with the current state of the product as well by using multiple individual elements from it

Here are a couple of things I think we should discuss before making the version 2.0 public

1. Provide the component multiple inputs so one can do the bulk transactions
2. Make sure CPE Editor is able to do bulk updates.

cc @alexed1 

Also, we should list the current limitations of the component to the community so they are aware of 1.x limitations

1. Essentially they are limited to only 50 records with save activity turned on because the component will hit the limit.
2. They can workaround to use multiple of this but this will make actions asynchronous.

jdayment:
Hey Guys I have been following I came up with the loop to overcome the 10 send limitation I am more than happy to help out where I can my apex is weak but my troubleshooting and thinking outside the box game is strong :)

jdpond:
Hey Josh @jdayment , your ears must be burning - we spent hours yesterday talking about how to integrate CPE with your needs and your name came up dozens of times.  https://unofficialsf.com/send-rich-email-with-the-new-sendhtmlemail-action/comment-page-2/#comment-1235

jdpond:
My current thinking is to create a Custom Metadata and allow you to assign and add to a parameter collection of that record type - eliminate the need for APEX altogether on your part.  Catch:  No idea if this would actually work yet.

msrivastav13:
@jdayment I am curious if you can use the v2.0 version that Alex have in this thread to see if you can still use it with your approach!

PR #453, opened 2020-08-13
sendBetterEmail - All relevant Collections (String and Object) now accessible for selection in CPE

jdpond:
Code was only looking in "lookups" in flows.  Should have also been looking in variables - for both objects (record collections) and strings (string collections).

Also fixed but that was associating tasks with targetId instead of WhatId

Next Steps:

1. Modify the documentation for renaming and update the 6 flow example videos(after your preliminary review) (JDP)
2. Create the sendBetterEmail Version 2.0.0.0 unmanaged package (AE)

Subsequent steps:

- Refactor code for quality and maintainability
- Introduce validations into CPE instead of action class
- Introduce flow-level bulkification (allow multiple request "bags" to be created using mdt record). Not sure if this one is needed

### DIFF
--- a/flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmail.cls
+++ b/flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmail.cls
@@ -1,7 +1,7 @@
 /**
 * @File Name			: SendBetterEmail.cls
  * @Description			: Uses Spring/Summer '20 EmailTemplate Object + ContentVersion with multi-lingual
- * @Credits				: Alex Edelstein, Jeremiah Dohn, etal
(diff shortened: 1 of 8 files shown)


jdpond:
Package installations for testing are available (the latter is just for upgrading existing environemnts.  There will be no future releases of sendHTMLEmail - deprecated in favor of sendBetterEmail.

- [sendBetterEmail(0.0.0.1)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t3h0000045qtgAAA)
- [sendHTMLEmail(2.0.1.0)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t3h0000045qtbAAA)

PR #455, opened 2020-08-17
Create fbc_ExtractMetadataStringFromZipfile.cls

ericrsmith35:
Additional Class for Flow Base Components.  This takes the Zipfile output from fbc_RetrieveMetadata and returns an extracted string and an escaped string of the metadata in the Zip file.

### DIFF
--- a/flow_process_components/FlowBaseComponents/force-app/main/default/classes/fbc_ExtractMetadataStringFromZipfile.cls
+++ b/flow_process_components/FlowBaseComponents/force-app/main/default/classes/fbc_ExtractMetadataStringFromZipfile.cls
@@ -1,0 +1,84 @@
+/**
+ * Eric Smith - August 2020
+ * 
+ *      Extract a single metadata file from the Zip file returned by the Retrieve
+ *      method of the fbc_TransferMetada LWC
+ * 
+ *      Input:  Zip File String
+ * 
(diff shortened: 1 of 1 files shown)


ericrsmith35:
Added Test Class

flow_process_components/FlowBaseComponents/force-app/main/default/classes/fbc_ExtractMetadataStringFromZipfile.cls
```diff
@@ -0,0 +1,84 @@
...
+            curResult.extractedFlowMetadata = unescapedString;
+            curResult.escapedFlowMetadata = escape(unescapedString);
+            responseWrapper.add(curResult);
+        }
+
+        return responseWrapper;
+
+    }
+
+    public static String unpackZipfile(String metadataZipfile) {
```
alexed1, line 32: is this the same unpack zipfile as the one in Retrieve Metadata? Can we DRY these up?

alexed1: would it possibly make sense to add the unpack capability to the RetrieveMetadata class? There's an argument that it's a fit...

Issue #462, opened 2020-08-27
QuickChoice FSC Bug - Picklist not sorting alphabetically


Xtremefaith:
I have a Global Value Picklist on my `OrderItem` object. 
![image](https://user-images.githubusercontent.com/1228550/91420531-ad6a5c80-e809-11ea-8e70-dce0871219d1.png)

The values are effectively brought into the picklist selection but they are in the order in which they were added rather then obeying the "Display values alphabetically" setting. I don't know of a setting that needs to be changed on the component to respect this so I assume this is a bug.

![image](https://user-images.githubusercontent.com/1228550/91420333-68462a80-e809-11ea-81a4-0af97bb5e47b.png)

ericrsmith35:
I've added an option to sort the Picklist labels.  See Pull Request #464

PR #464, opened 2020-08-27
QuickChoice Picklist Sort Option Added

ericrsmith35:
Added an option to sort the Picklist labels.  See Issue #462


ericrsmith35:
@alexed1 Are you able to create a new install package for this?

PR #465, opened 2020-08-28
Fixed regex to not strip __ from field names

adwhite75:
Noticed the field name was getting truncated if it had __c in it so I found the issue in the regex that cleans up the field names


alexed1:
Thanks for the fix, Adam!

PR #466, opened 2020-08-29
support for String.Split, Strings.Join, Strings.Unique

ChuckJonas:
I started writing something similar, and then I found this project (glad I did because it does a bunch of what I was going to do plus TON more.  Nice work!!!!)

Anyways, I had a couple invocable that I didn't see in here (incoming via this PR and one other).

- `String.Join`
- `String.split`: I've found this really nice as it's a easy way to init collects of strings
- `String.unique`: filters a collection of string, removing duplicates 

I wasn't sure the best place to put them, so let me know if I need to move these or create a new folder.  Also, I can split out the tests to individual files

If you don't think they are a valuable addition, no worries!

### DIFF
--- a/flow_action_components/StringChangers/force-app/main/default/classes/StringSplit.cls
+++ b/flow_action_components/StringChangers/force-app/main/default/classes/StringSplit.cls
@@ -1,0 +1,39 @@
+global class StringSplit {
+  @InvocableMethod(
+    label='Splits a string in an array of strings'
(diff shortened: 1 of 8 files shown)


ChuckJonas:
Sure, happy to.  sent you an email to the address listed in your github profile.  

Before you generate the package, I actually need to clean these up a bit (sorry should have taken time to do so before submitting, but wasn't sure if they fit in).

Should be able to push the needed changes tomorrow

PR #467, opened 2020-08-29
-added support for adding and removing multipicklist values

ChuckJonas:
2nd PR from the collection.

I couldn't tell if this capability already exists... It looks like maybe the ability to Add to a multi-picklist was part of the existing "picklist" but at the very least it seemed like remove wasn't implemented.

I can make changes as needed.  Cheers

### DIFF
--- a/flow_action_components/MultipicklistValues/force-app/main/default/classes/MultiPicklistAdd.cls
+++ b/flow_action_components/MultipicklistValues/force-app/main/default/classes/MultiPicklistAdd.cls
@@ -1,0 +1,49 @@
+global class MultiPicklistAdd {
+  @InvocableMethod(
+    label='Add Multi-Picklist Val(s)'
+    description='Given A multi-picklist value (as string), will add the target the value(s) from it and return a new picklist value (as string)'
+    category='Picklist'
+  )
+  global static List<String> addPicklistVal(
+    List<MultiPicklistAddParams> params
+  ) {
+    List<String> results = new List<String>();
+    for (MultiPicklistAddParams p : params) {
+      String pickVals = p.picklistValues;
+      String[] currentVals = pickVals.split(';');
+      for (String target : p.target.split(';')) {
+        target = target.trim();
+        //check if the current value is already in the picklist (case-insensitive).  If not, add it
+        Boolean found = false;
+        for (String s : currentVals) {
+          if (s.toLowerCase() == target.toLowerCase()) {
+            found = true;
+            break;
+          }
+        }
+
+        if (!found) {
+          pickVals += ';' + target;
+        }
+      }
+      results.add(pickVals);
+    }
+    return results;
+  }
+
+  global class MultiPicklistAddParams {
+    @InvocableVariable(
+      label='Target Value'
+      description='The Value(s) to Add. Supports multi, ";" separated values'
+      required=true
+    )
+    global String target;
+
+    @InvocableVariable(
+      label='Picklist Values'
+      description='The current value of the multi-picklist'
+      required=false
+    )
+    global String picklistValues;
+  }
+}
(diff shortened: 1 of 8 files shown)


ChuckJonas:
Yea... so these a definitely a bit more "targeted" to just updating Multi-picklists in a single action.  You provide the picklist value and values to add or remove and it handles the rest (dealing with duplicates, the value not actually being there, etc).

Actually now that I look at it... the method you linked is very similar to the `Split` and `Join` functionality I submitted in the 466 & 468 PR.  Just with the `Convert CSV` ones you can't actually specify the delimiter...

Not sure if it makes more sense for me to instead submit a PR to update those methods (would have to maintain backwards compatibility and the naming would be a bit odd).

**Here's the use-case I was going to demo for those:**

> You've been tasked with de-duplicating a field which contains a list of emails, seperated by the `|` character.
> Example Input:
> `john@Doe.com|jane@example.com|JOHN@doe.com|jane@example.com`

Flow_Builder

https___inspiration-speed-7056-dev-ed--c_visualforce_com_flow_runtime_apexp_flow__debug_true_flowVersionId_301R0000000GfUw_flowDevName_Split_Join_Unique_demo_debugInputs__5B_7B_name__3A_InputString__2C_collection__3Afalse_2C_type__3A_String

flow_action_components/MultipicklistValues/force-app/main/default/classes/MultiPicklistAdd.cls
```diff
@@ -0,0 +1,49 @@
+global class MultiPicklistAdd {
+  @InvocableMethod(
+    label='Add Multi-Picklist Val(s)'
+    description='Given A multi-picklist value (as string), will add the target the value(s) from it and return a new picklist value (as string)'
+    category='Picklist'
+  )
+  global static List<String> addPicklistVal(
+    List<MultiPicklistAddParams> params
```
alexed1, line 8: Two comments on your approach to inputs:

1) the standard pattern we use is to define Request and Response classes and pass those in and out of the invocable action, instead of directly passing data like you do here. This enables bulkification to work automatically: if someone triggers a flow on a batch of 40 records and your action is in that flow, it will not automatically and properly bulkify your current code because it doesn't have the empty outer list that Request and Response provide. For examples, see the String Changers stuff. While this isn't a breaking problem, I'm motivated to keep consistency unless there's a good reason.

You're making your input an Apex-Defined type that contains two strings. That's ok, but in this case, I would think you'd just want to put the two strings independently into the 'Request' class. That way they show up as individual strings in Flow Builder.

alexed1 (changes requested):

alexed1:
I think that flow looks like a great demo for the StringChanger improvements and I encourage you to put it in the posts I just requested.

Back to your Multiselect PR: We've neglected utilities for multiselect picklists and I think your two actions would be a good nucleus for a new MultiselectUtils package of actions. I left a couple of specific comments about the code.

flow_action_components/MultipicklistValues/force-app/main/default/classes/MultiPicklistAdd.cls
alexed1: Can you add as comments in the code some examples of inputs and the resulting outputs?

alexed1:
Hi Chuck. Just checking in on this. I'd love to get this pulled in
But was wondering if you could make a few changes to make the code style more consistent. I left a few comments￼

flow_action_components/MultipicklistValues/force-app/main/default/classes/MultiPicklistAdd.cls
ChuckJonas: I'll take a look at the other string changers and refactor to make.  I'll try to get around to it this week (as well as wrap up the split/join/unique stuff)

alexed1:
I'm going to close this for now, but feel free to reopen it if you get some time.

PR #468, opened 2020-08-31
Strings.Join, Split Unique (Cleanup PR)

ChuckJonas:
- cleaned up code & labels
- moved test to individual files
- added demo flow


alexed1 (approved):
I see you're using some regex. Most of our users actually either have never used regex or are intimidated by it (with good reason), so make sure you provide several examples of how to do the most common kind of delimiting (like delimit on a space, a comma or a semicolon) and then maybe a few examples that show off the power of what regex can really do.

alexed1:
Created an unofficialsf account for your Callaway address. A 'Documentation Post' is the blog post that is kept up to date as an extension evolves, with all of the documentation and the latest install link.  Here's a good example of a good documentation blog post: https://unofficialsf.com/quick-choice-improved-picklists-and-radio-buttons-for-flow/  

String Changers never really got a proper Documentation Post. What we have is here: https://unofficialsf.com/flow-actions-for-strings-url-encoding-and-number-formatting/ . So what I'd like to suggest you do is put your content in basically two places: a brand new Announcement Post that announces the new stuff you've built, and also modify [the existing post ](https://unofficialsf.com/flow-actions-for-strings-url-encoding-and-number-formatting/) so that it has the complete up to date documentation. If you look at the bottom of https://unofficialsf.com/detectandlaunch-1-1-launch-_screen_-flows-modally-when-records-are-edited-or-deleted/ you can see an example of how an Announcement Post refers people over to the permanent Documentation Post.

PR #475, opened 2020-09-01
Add DedupeSObjectCollectionInvocable

agentgill:
Super simple dedupe of a generic sobject collection. There are a couple of ways to handle this directly in flow, but involves looping and/and multiple assignment elements. This is just sobject collection in / out, job done.

### DIFF
--- a/flow_action_components/DedupeSObjectCollectionInvocable/.eslintignore
+++ b/flow_action_components/DedupeSObjectCollectionInvocable/.eslintignore
@@ -1,0 +1,15 @@
+**/lwc/**/*.css
+**/lwc/**/*.html
+**/lwc/**/*.json
+**/lwc/**/*.svg
+**/lwc/**/*.xml
+**/aura/**/*.auradoc
+**/aura/**/*.cmp
+**/aura/**/*.css
+**/aura/**/*.design
+**/aura/**/*.evt
+**/aura/**/*.json
+**/aura/**/*.svg
+**/aura/**/*.tokens
+**/aura/**/*.xml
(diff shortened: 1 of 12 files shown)


adwhite75:
@alexed1  I am happy to upload the source for that in here and we can make adjustments as needed - just let me know!

agentgill:
Sorry got slammed after issuing this one. Would be good to see / have @adwhite75 upload his code if his code is very close. It feels like the pattern to handle dedupe based on the field would be different from the approach I used, not sure. The use case I was addressing was performing multiple get records across the same object type and feeding the collection into a data table - I was looking for a simple way to dedupe list collection.

alexed1:
Great! go ahead and create a PR for the original dedupe, Adam, and let's compare the two

adwhite75:
@agentgill  sorry I am behind on getting the code in here! Havent forgotten

alexed1:
no problem. looking forward to it.

adwhite75:
@agentgill  here is the code: https://github.com/adwhite75/LightningFlowComponents/tree/RecalculateFormulas/flow_action_components/DedupeRecordCollection/force-app/main/default/classes

I didnt get around to making a test class for it yet

alexed1:
@agentgill let me know if you'd like to take a look at this. The question is mainly: can this be a single action instead of 2? It's in our interest to avoid fragmentation. However, if you don't have time, totally understand, and really appreciate your contribution.

alexed1:
I'm going to close this for now, but we can reopen it anytime

PR #505, opened 2020-10-05
ADD node.js server for RPA

andkraiev:


alexed1 (approved):
This is some exploratory work in building local actions that can manipulate the local computer by sending commands to a localhost web server. in theory, this would allow flow to do RPA things and also be a UI for SFDX commands.

PR #506, opened 2020-10-05
QuickChoiceFSC 

ChewinChaz:


ericrsmith35:
@alexed1 How are you going to handle updates to this and other components that also are part of the Flow Base Components?

alexed1:
@ChewinChaz  Thank you for contributing changes to the project! Can you give us a description/writeup of what you've done? Can you record a little animated gif or a screen cast that demonstrates it? 

@ericrsmith35  I think we want to just make the changes to the fbc version and in fact consider removing the 'standalone' versions or at least making them less visible. What do you think? 

@ChewinChaz we just moved this component into a library of base components here: https://github.com/alexed1/LightningFlowComponents/tree/master/flow_process_components/FlowBaseComponents so that's probably where we're going to want to have your changes.

ericrsmith35:
@alexed1 I'd already gone ahead yesterday and changed the USF pages for QuickChoice and DualListBox to point to the FlowBaseComponents page for installation.

ChewinChaz:
@alexed1 Very strange, I did a short write up and it seems to have vanished!  Apologies, hopefully this one doesn't do the same:

I have added a checkbox labeled "Button Mode" to the QuickChoice CPE that, when checked, automatically advances the user to the next screen.  I have found it to be exceptionally useful when the user is required to make several quick choices and additional processing is done in between each selection to determine the next choices to be displayed.  In my use case, I have created a flow that allows project supervisors to clock their crew in and out via their mobile phones which allows me to take a timestamp and a GPS snapshot with a click of a button.

Screen #1 - Home
QC1: **Punch Time Card**
QC2: Edit Crew
QC3: View Work Order

[Processing]
Get Current Crew, Get Past Daily Reports, etc.

Screen #2 - Clock in or out?
QC1: Clock In
QC2: **Clock Out**

[Processing]
Get Open Time Cards for Current Crew, Get previous time card punch records

Screen #3 - Clock out type
QC1: Regular Time
QC2: Certified Time
QC3: Lunch Break

[Processing]
Action: Retrieve lat/lng, local time, server time, Update Time Card, Update Crew status, etc.

The supervisors have to do this for every location, every project, every day; which means they may run this flow over a dozen times throughout their workday.  "Button Mode" cuts the number of clicks in half by removing the need to make a selection and then click Next to advance to the next screen and eliminates the need to preemptively execute flow elements not related to the current choice so conditional visibility can be used.

I have found it especially useful in the mobile app for making quick selections as you don't need to add several inputs to a single screen, forcing the user to scroll and advance next manually.  It can also be used as a nice replacement for the lightning-button component since the number of "buttons" are dynamic and looks much cleaner as the size of each is larger and the width is uniform rather than being determined by the length of the label.

lightning-button group
image

QuickChoice button group
image

ChewinChaz:
qcButton

ChewinChaz:
@alexed1 @ericrsmith35 I am fairly new to salesforce development and very new to lwc's, so if you do decide you wish to implement my addition, please check over my code first!  It works perfectly from my tests but could certainly be missing something or have syntax that could use improvement.  

I am also relatively new to collaborating through github unfortunately, could y'all expound on your previous comments for me concerning the pull request?  I can certainly add the new commits to the QuickChoice component in the flowbasecomponents repository and issue another request if needed.

alexed1:
This is really great. Yes, please do these things next:

1) apply all the code changes you just applied to the copy of quick choice and quick choice cpe in the flow base components location here: https://github.com/alexed1/LightningFlowComponents/tree/master/flow_process_components/FlowBaseComponents/force-app/main/default/lwc
the way I'd do this: 

a) in your desktop file system create a copy of all of the project files in the quick choice and quick choice cpe components. This code should contain your changes. basically just copy and paste onto your desktop for temporary safe keeping
b) run the git revert command and revert back to the last commit before your changes. this will remove the changes you made to the deprecated quick choice location
c) copy your modified project files to the new location (shown in #1 above), overwriting files
d) create a new pull request
verify that the same set of proposed changes is being made to the same files, just in a different location.

We'll switch over to that pull request and essentially abandon this one. 

2) I'm thinking we maybe call this InstantNavigation Mode, because 'Button' doesn't really differentiate. 

3) I'd like you to expand your description from above into a blog post. If you're good with that, let me know your email of choice and I'll give you author rights on unofficialsf.com. When you expand it into the post, provide step by step instructions if necessary. I'm thinking about externalities like hiding the footer. I saw that you have navigation components in your gif. Are they required? if so doc that. When you're ready, save it to draft and let me know here and I'll do an edit pass


@ericrsmith35 , anything to add?

ChewinChaz:
@alexed1 Can do, stoked you like it!  I just called it button mode as it essentially changes the component from a visual picklist to a a visual button group, but I'm fine with calling it whatever.  To that point though, is it possible to create help dialogues on CPE inputs as can be done with aura components?  I feel that whatever the label, a short description seems necessary.   

My github email is sirchaseman@gmail.com so thats probably the best choice

ericrsmith35:
It's a nice addition.  I originally set it up that way but pulled it
because of wanting to support users requiring adaptive resources such as
screen readers.  I do think there are some great use cases however for
having the click navigate to the next node in the flow.  The documentation
should probably point out that there will be a certain subset of users
unable to interact with flows created with this option enabled.

Eric

ChewinChaz:
Migrated to fbc version of Quick Choice and issued new pull request

https://github.com/alexed1/LightningFlowComponents/pull/507

PR #512, opened 2020-10-15
Initial Commit Get Opportunity Partners Action

Tamarer:
@alexed1 , this is for your review, thanks!

### DIFF
--- a/flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
+++ b/flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
@@ -1,0 +1,132 @@
+/**
+* @author       : Tamar Erlich
+* @date         : October 07, 2020
+* @description  : Invocable method that given opportunityId and accountId, returns a list of opportunity partners to display on opportunity plan record page
+* Note          : This class is called by the Flow.
+* *************************************************
+*       <runTest>GetOpportunityPartnersActionTest<runTest>
+* *************************************************
+* @History
+* -------
+* VERSION | AUTHOR                | DATE               | DESCRIPTION
+* 1.0     | Tamar Erlich          | October 07, 2020   | Created
+**/
+global with sharing class GetOpportunityPartnersAction {
+
+    // Attributes passed in from the Flow
+    global class Requests {
+
+        @InvocableVariable(label='OpportunityId' required=true)
+        global String opportunityId;
+        
+        @InvocableVariable(label='AccountId' required=true)
+        global String accountId;
+    }
+
+    // Attributes passed back to the Flow
+    global class Results {
+
+        public Results() {
+            
+        }
+        @InvocableVariable
+        global String errors;
+
+        @InvocableVariable
+        global String successful;
+
+        @InvocableVariable
+        public List<OpportunityPartnersWrapper> outputCollection;
+
+        @InvocableVariable
+        public String outputString;
+        
+    }
+
+    // custom exception class
+    global class InvocableActionException extends Exception {}
+
+    // Expose this Action to the Flow
+    @InvocableMethod
+        global static List <Results> get(List<Requests> requestList) {
+            // initialize variables
+            Results response = new Results();
+            List<Results> responseWrapper = new List<Results>();
+            String errors;
+            String success = 'true';
+            String stringOutput;
+            String accountId = requestList[0].accountId;
+            String opportunityId = requestList[0].opportunityId;
+            List<OpportunityPartner> opportunityPartners = new List<OpportunityPartner>();
+            List<OpportunityPartnersWrapper> partners = new List<OpportunityPartnersWrapper>();
+
+// BEGIN APEX ACTION PROCESSING LOGIC
+            try {
+
+                if (accountId == null || opportunityId == null){
+
+                    throw new InvocableActionException ('No Account or Opportunity ID');
+
+                }
+
+                if (!accountId.startsWith('001')|| !opportunityId.startsWith('006')){
+
+                    throw new InvocableActionException ('Invalid Account or Opportunity ID');
+
+                }
+
+                if (accountId != null && opportunityId != null){
+                   
+                opportunityPartners = [SELECT AccountTo.Name, IsPrimary, Role 
+                                        FROM OpportunityPartner 
+                                        WHERE OpportunityId =: opportunityId and AccountToId !=: accountId];
+                
+                }
+
+                // populate the Apex defined wrapper type with opportunity partner information
+                for (opportunityPartner op : opportunityPartners) {
+
+                    OpportunityPartnersWrapper opw = new OpportunityPartnersWrapper();
+
+                    opw.accountName = op.AccountTo.Name;
+                    opw.partnerRole = op.Role;
+                    opw.isPrimary = op.IsPrimary;
+                    opw.accountLink = '/'+op.AccountToId;
+
+                    partners.add(opw);
+
+                }
+
+                // Convert Record Collection to Serialized String
+                stringOutput = JSON.serialize(partners);
+
+            } catch (InvocableActionException e){
+                System.debug ('exception occured: ' + e.getMessage());
+                
+                errors = e.getMessage();
+                success = 'false';
+            } catch (exception ex){
+                System.debug ('exception occured: ' + ex.getMessage());
+                
+                errors = ex.getMessage();
+                success = 'false';
+            }
+// END APEX ACTION PROCESSING LOGIC
+
+            // Prepare the response to send back to the Flow
+            // Set Output Values
+            response.errors = errors;
+            response.successful = success;
+            response.outputCollection = partners;
+            response.outputString = stringOutput;
+            
+            responseWrapper.add(response);
+    
+            // Return values back to the Flow
+            return responseWrapper;
+
+        }
+
+
+
+}
--- a/flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
+++ b/flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
@@ -1,133 +1,138 @@
 /**
-* @author       : Tamar Erlich
-* @date         : October 07, 2020
-* @description  : Invocable method that given opportunityId and accountId, returns a list of opportunity partners to display on opportunity plan record page
-* Note          : This class is called by the Flow.
-* *************************************************
-*       <runTest>GetOpportunityPartnersActionTest<runTest>
-* *************************************************
-* @History
-* -------
-* VERSION | AUTHOR                | DATE               | DESCRIPTION
-* 1.0     | Tamar Erlich          | October 07, 2020   | Created
-**/
+ * @author       : Tamar Erlich
+ * @date         : October 07, 2020
+ * @description  : Invocable method that given opportunityId and accountId, returns a list of opportunity partners to display on opportunity plan record page
+ * Note          : This class is called by the Flow.
+ * *************************************************
+ *       <runTest>GetOpportunityPartnersActionTest<runTest>
+ * *************************************************
+ * @History
+ * -------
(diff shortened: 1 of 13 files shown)


flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
```diff
@@ -0,0 +1,132 @@
...
+* -------
+* VERSION | AUTHOR                | DATE               | DESCRIPTION
+* 1.0     | Tamar Erlich          | October 07, 2020   | Created
+**/
+global with sharing class GetOpportunityPartnersAction {
+
+    // Attributes passed in from the Flow
+    global class Requests {
+
+        @InvocableVariable(label='OpportunityId' required=true)
```
alexed1, line 19: When writing labels we generally try to make use of spaces and emphasize readability, rather than just repeat the name of the underlying unique name.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
```diff
@@ -0,0 +1,132 @@
...
+* *************************************************
+* @History
+* -------
+* VERSION | AUTHOR                | DATE               | DESCRIPTION
+* 1.0     | Tamar Erlich          | October 07, 2020   | Created
+**/
+global with sharing class GetOpportunityPartnersAction {
+
+    // Attributes passed in from the Flow
+    global class Requests {
```
alexed1, line 17: If you look at the other actions, you'll see that our convention is to start with the invocable method and put the Request and Response classes at the bottom. It's not necessary, but the consistency makes things a little easier to work with.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
```diff
@@ -0,0 +1,132 @@
...
+        global String opportunityId;
+        
+        @InvocableVariable(label='AccountId' required=true)
+        global String accountId;
+    }
+
+    // Attributes passed back to the Flow
+    global class Results {
+
+        public Results() {
```
alexed1, line 29: generally, erase empty constructors like this to keep your code tight.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
```diff
@@ -0,0 +1,132 @@
...
+            
+        }
+        @InvocableVariable
+        global String errors;
+
+        @InvocableVariable
+        global String successful;
```
alexed1, line 36: another general convention is that if you're returning something that only has two states, such as success or failure, use a Boolean type, perhaps calling it isSuccess. If you have multiple values, then a String is a good solution.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/OpportunityPartnersWrapper.cls
```diff
@@ -0,0 +1,45 @@
...
+* Note          : This class is called by the GetOpportunityPartnersAction.
+* *************************************************
+*       <runTest><runTest>
+* *************************************************
+* @History
+* -------
+* VERSION | AUTHOR                | DATE               | DESCRIPTION
+* 1.0     | Tamar Erlich          | October 07, 2020   | Created
+**/
+public with sharing class OpportunityPartnersWrapper {
```
alexed1, line 14: 'Wrapper' generally is used to refer to action code that calls something and then handles its outputs, and is designed specifically to enhance the thing it's wrapping. I don't use it for data structures like this. Here, I'd probably go with OpportunityPartnerSet or OpportunityPartnerDescription

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/OpportunityPartnersWrapper.cls
```diff
@@ -0,0 +1,45 @@
+/**
+* @author       : Tamar Erlich
+* @date         : October 07, 2020
+* @description  : Wrapper class for opportunity partners
+* Note          : This class is called by the GetOpportunityPartnersAction.
```
alexed1, line 5: this is really good documentation. If someone stumbles across this class, you're giving them crucial context.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
```diff
@@ -0,0 +1,132 @@
...
+            List<Results> responseWrapper = new List<Results>();
+            String errors;
+            String success = 'true';
+            String stringOutput;
+            String accountId = requestList[0].accountId;
+            String opportunityId = requestList[0].opportunityId;
+            List<OpportunityPartner> opportunityPartners = new List<OpportunityPartner>();
+            List<OpportunityPartnersWrapper> partners = new List<OpportunityPartnersWrapper>();
+
+// BEGIN APEX ACTION PROCESSING LOGIC
```
alexed1, line 63: this kind of comment doesn't help, because it's self-evident, so I'd leave it out.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
```diff
@@ -0,0 +1,132 @@
...
+    // Expose this Action to the Flow
+    @InvocableMethod
+        global static List <Results> get(List<Requests> requestList) {
+            // initialize variables
+            Results response = new Results();
+            List<Results> responseWrapper = new List<Results>();
+            String errors;
+            String success = 'true';
+            String stringOutput;
+            String accountId = requestList[0].accountId;
```
alexed1, line 58: Most actions created through 2019 used this [0] approach, but it was pointed out that they lost the ability to work in batch scenarios. So these days, we instead use the for loop approach that you can see here: https://github.com/alexed1/LightningFlowComponents/blob/master/flow_action_components/CollectionProcessors/force-app/main/default/classes/CopyCollection.cls

Once you do it once or twice it becomes very natural.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
```diff
@@ -0,0 +1,132 @@
...
+            String opportunityId = requestList[0].opportunityId;
+            List<OpportunityPartner> opportunityPartners = new List<OpportunityPartner>();
+            List<OpportunityPartnersWrapper> partners = new List<OpportunityPartnersWrapper>();
```
alexed1, line 61: not a good variable name. too general. For a data type like this, I'd use something like oppPartnerWrapperSet or oppPartnerWrappers.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
```diff
@@ -0,0 +1,132 @@
...
+
+// BEGIN APEX ACTION PROCESSING LOGIC
+            try {
+
+                if (accountId == null || opportunityId == null){
+
+                    throw new InvocableActionException ('No Account or Opportunity ID');
```
alexed1, line 68: I love the way you're taking the time to do input validation. I'd put a little more detail into your error text. Remember that the user may have very little context. So for this, I'd go with something like "When using the GetOpportunityPartners action, you need to provide BOTH an Account Id AND an Opportunity Id"

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersActionTest.cls
```diff
@@ -0,0 +1,169 @@
...
+        for (Integer j = 0; j < 1; j++) {
+
+            Account a = new Account(
+                Name = 'email' + j + '.com'
+                );
+
+            accounts.add(a);
+        }
+        
+        insert accounts;
```
alexed1, line 99: fyi we have run into some problems where people try to install these actions, but on their orgs, there are some additional fields on the object (here, Account) that have to exist in order to create a record. This causes their test to fail because it can't create an Account. 

The main solution for this is to distribute as a managed package, which doesn't run tests. There's overhead involved in doing that, though, because you need to create a separate dev edition org for each managed package. 


I don't have any suggestions here. Just letting you know it's an issue. It's a good thing to have tests.

alexed1 (changes requested):
Solid code, Tamar. I made a few suggestions.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
Tamarer: I tried to add the for loop but then I got a SOQL query inside a loop error

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
Tamarer: I opted for keeping the string and adding more granular error messages

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/OpportunityPartnersWrapper.cls
Tamarer: Isn't this what e are doing here? taking the query output from the action and enhancing it by adding an account name and account link?
I let the name as is because modifying it now , will require modifying the blog images and the flow as well.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersActionTest.cls
Tamarer: I have encountered this and had to rewrite the test class for datatable in my org to accommodate for required account field :)

Tamarer:
Thank you for all your feedback @alexed1 I tried to make the requested changes.

flow_action_components/GetOpportunityPartners/force-app/main/default/classes/GetOpportunityPartnersAction.cls
alexed1: Can you show me the code where you got the error? We really don't want to publish new actions that aren't bulkified....

Tamarer: Hi @alexed1 , I think I was able to bulkify the action code (and learned how to use maps along the way :) ) but I would like to understand why it is relevant here since we are ever only passing a single opportunityId into the action?

alexed1: Most bulkification scenarios start with 'suppose someone uploads a batch of 50 opportunities, and that triggers a flow that has this action?'  Flow will try to send all 50 to this action. It's good for performance.

Tamarer: Got your point, thanks for explaining.
Is there anything else I need to refactor in this action?

alexed1:
just go through the comments I left and decide if you want to make any more
changes

Alex Edelstein
VP Product Management - Automation, Orchestration and Decisioning
Salesforce

<https://smart.salesforce.com/sig/aedelstein//us_mb/default/link.html>

Tamarer:
Hi @alexed1 I have made most of the required changes (except renaming Apex object and successful string) and bulkified the code. For my original use case, I do not have any more changes to make. Thanks for your review and comments, Tamar.

alexed1:
Great stuff, Tamar. Go ahead and publish a post about it and add a link to the Flow Actions table. (while you're doing that, can you check to see if recent posts about Actions and Screen Components have  resulted in links in the appropriate tables? )

PR #515, opened 2020-10-18
Flow Combobox Support for Autocomplete and Help Text

egolden-tc:
This PR includes two new public attributes on Flow Combobox:

autocomplete - this maps to the standard autocomplete field on lightning-input.  By default it is set to "off".  This ensures browser autocomplete dropdown doesn't cover or conflict with the combobox options.  It can be changed to "on" to revert to the original behavior.

field-level-help - this maps to the standard field-level-help on the lightning-input.  If a value is provided it will show as help text on the combobox to give users context on what they are selecting and/or why.

The only default behavior change is the disabling of the autocomplete. This is probably desirable but the behavior can be reverted to the previous default per the above options.

### DIFF
--- a/flow_process_components/FlowBaseComponents/force-app/main/default/lwc/fbc_flowCombobox/fbc_flowCombobox.html
+++ b/flow_process_components/FlowBaseComponents/force-app/main/default/lwc/fbc_flowCombobox/fbc_flowCombobox.html
@@ -10,7 +10,7 @@
                                  role="none" style="height: 55px;">
                                 <template if:false={displayPill}>
                                     <div onkeyup={handleSearchKeyUp}>
-                                        <label class="slds-form-element__label custom-width-full" onclick={toggleMenu}>{label}</label>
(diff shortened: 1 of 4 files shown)


egolden-tc:
@alexed1 Let me know if more context or details are needed.  These updates are minor but were necessary for the custom property editors I'm working on to behave consistently.

egolden-tc:
Upon further review it looks like the label markup is different in this version of the combobox which is causing some alignment issues.  Maybe hold off on reviewing this until I have a chance to update my help text markup/css.  Sorry for the delay!!

egolden-tc:
@alexed1 I have fixed the field level help markup issue.  This is ready to review when you have time. 
 Thanks!

alexed1:
This looks great, but should go in a different location because we've basically abandoned FlowBaseComponents in favor of a new pair of projects called FlowActionBasePack and FlowScreenComponentBasePack. See https://unofficialsf.com/introducing-flow-basepacks/

Can you instead add this code to the version of FlowCombobox at https://github.com/alexed1/LightningFlowComponents/tree/master/flow_screen_components/FlowScreenComponentsBasePack/force-app/main/default/lwc/fbc_flowCombobox ? This is the one we're going to be using, going forward. I.e. all packages that use flowcombobox will be built with this codeline.

ericrsmith35:
This has now been included in the standard fsc_flowCombobox

PR #517, opened 2020-10-28
ChatSessionManager

ehretg:
Adding two new components:

- **Screen component** ChatSessionManagerSC

- **Action** ChatSessionManagerLA


flow_action_components/ChatSessionManagerLA/ChatSessionManagerLA.cmp
```diff
@@ -0,0 +1,7 @@
+<aura:component implements="lightning:availableForFlowActions" access="global">
```
alexed1, line 1: Since this is a Local Action, I suggest you name it ManageChatSession instead of ChatSessionManagerLA. We like to name actions starting with a verb.

flow_action_components/ChatSessionManagerLA/ChatSessionManagerLAHelper.js
```diff
@@ -0,0 +1,80 @@
+({
+    sendMessage: function(component, event) {
+        var conversationKit = component.find("conversationKit");
+        var sessionId = component.get("v.sessionId");
+        
+        if (sessionId != null && sessionId != '') {
```
alexed1, line 6: There's a lot of repeated code in these functions. It would be nice to DRY it up...

flow_screen_components/ChatSessionManagerSC/ChatSessionManagerSCHelper.js
```diff
@@ -0,0 +1,80 @@
+({
```
alexed1, line 1: Is this exactly the same code as ChatSessionManagerLAHelper.js ? If that's the case, I think that I would change things so that the action and screen component live in the same project, and they both import the same javascript file, so that you only have to have one version of this code. ...

alexed1: I think I would consolidate it in the flow_screen_components/ChatSessionManager directory. Then we can package it in a single package with both the action and the screen component.

alexed1 (changes requested):

flow_action_components/ChatSessionManagerLA/ChatSessionManagerLA.cmp
ehretg: I will create a new pull request with a new name for the action component : ManageChatSession

flow_action_components/ChatSessionManagerLA/ChatSessionManagerLAHelper.js
ehretg: I have changed the code and added the repeated code lines in the controller instead of helper

flow_screen_components/ChatSessionManagerSC/ChatSessionManagerSCHelper.js
ehretg: My initial objective was to have one component. What I noticed and lead me to create two separate components:

- With one component you should have two methods in the controller "init" (for the screen component) and "invoke" (for the local action)

- When using the component as an Action both methods "init" and "invoke" are called
So I split  the component in two separate components

Issue #481, opened 2020-09-03
sendBetterEmail CPE - allow flow formulas (string)  to be selected in flow combo boxes


jdpond:
For recipients, allow formulas to be selected if they are of type string in the combobox drop downs

Issue #450, opened 2020-08-11
sendBetterEmail - Update Examples to reflect CPE Labels and Functionality


jdpond:
All 6 of the examples could use a cleanup so that the labels are in sych with the CPE.  At the same time look at adding the saveActivity/saveTask split and using more of the name functions rather than the Ids

jdpond:
Concurrently update the example documentation

jdpond:
Both the examples and documentation have been upgraded

Issue #449, opened 2020-08-11
Mass Email Bug


shasta824-deli:
The setTargetObjectIds section of the mass send code had an error when setting the whatIds.  It was resetting the targetobjectIds to the whatId list.  This was on line 119 of SendHTMLEmail version I got from the 1.33.2 version.

Sorry I am not sure how to notify or go through GitHub to announce these.  

I really like that you can send mass emails getting around the 10 email limit per call to the single message version.

jdpond:
Could you retry this with https://login.salesforce.com/packaging/installPackage.apexp?p0=04t3h0000045qiMAAQ

shasta824-deli:
Yes.  I am still documenting the other changes I needed to make to allow
for Document Attachments vs content document links.  Then I will uninstall
and give this a try.

Thanks and again I apologize that I do not know my way around GitHub very
well.

shasta824-deli:
I got the same error. "EXCEPTION_THROWN [349]|SendHTMLEmail.InvocableActionException: No Messages were sent.  First Error: Error Code: INVALID_ID_FIELD - Only Users, Contacts, Leads, and Person objects are allowed for targetObjectIds."

Looks like the same line of code(now line 123) is still setting the target object instead of the what ids.

I did like the lightning component interface for the action. However it did not include parameters for attachments.  Which also don't include document attachments instead of just document links.  I added a parameter for this type of attachment.  Not sure if you would be interested in adding that to the code or not.  I had to do it because Salesforce CPQ still builds their Quote documents as the old version document instead of the new lightning files.

jdpond:
This is a problem with Collections in CPE - working on it now

jdpond:
shasta, I just found the problem in the sendHTMLEmailCPE for collections too.  Will release them both in a couple hours.

jdpond:
Fixed in latest pull request - test as soon as the package comes out and we can close.  THANK YOU!

Issue #351, opened 2020-05-03
sendHTMLEmail - Address bulkification consistent with Apex Gov Limits


jdpond:
sendHTMLEmail uses both the sendEmail and SingleEmailMessage method. Enhance to bulkify effectively to achieve the 10x10 maximum allowed under [Apex Governor Limits](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_gov_limits.htm)

Per-Transaction Apex Limits
Total number of sendEmail methods allowed 10
Per-Transaction Certified Managed Package Limits
Total number of sendEmail methods allowed 110

- The single and mass email limits don't take unique addresses into account. For example, if you have johndoe@example.com in your email 10 times, that counts as 10 against the limit.
- You can send an unlimited amount of email to your org’s internal users, which includes portal users.
- You can send mass emails only to contacts, person accounts, leads, and your org’s internal users.
- In Developer Edition orgs and orgs evaluating Salesforce during a trial period, you can send mass email to no more than 10 external email addresses per day. This lower limit doesn’t apply if your org was created before the Winter ’12 release and already had mass email enabled with a higher limit. Additionally, your org can send single emails to a maximum of 15 email addresses per day.

jdpond:
@alexed1 In bulkification, any given message may fail.  What would you think if we eliminate (generally) the use of the "throw new InvocableActionException" and expect the flow user to check Response.isSuccess.  This would allow (for example) 5 of 7 to sent and provide for consistent handling of errors.

alexed1:
sounds like a good idea

jdpond:
OK, will do.  I've done the bulkification - sorry about yesterday afternoon's brain snarl.  I'm going to also add a condition to "throw new InvocableActionException" if all sends fail with the first failure message.

jdpond:
Added the bulkification - works as follows:
1) All requests are mapped for responses in serial order (e.g., if there are 5 requests, map of 1-5 responses created) using a new private class that includes the response and various information that may also be needed for post send processing (e.g., activity information such as toAddresses)
2) If a request has an error in one of the error checks, it is noted in the mapped, but not included in the  sendemail collection
3) Upon response, SendEmailResult is mapped back to the responses maps.
4) For successful sends, activity tasks are created if requested through parameters
5) If an error occurs in task activity creation, it is recorded in the error message, but isSuccess remains true
6) Response list is created from mapped requests
[...]

PR #535, opened 2020-11-29
Upgraded to Flow BasePacks, fixed multiple bulkification errors, updated examples, Bulk Email

jdpond:
1. Attachments have ONLY been tested with CDL - not sure old attachment work.
2. Didn't test organization wide defaults
3. Unable to figure out how to default the signature to true if not previously specified since when a checkbox is changed, it goes to null rather than false (boolean).  Will work on this bug later.
4. Documentation is out of date - will work on this.

Resolves: 
#481 allow flow formulas (string) to be selected in flow combo boxes
#351 sendHTMLEmail - Address bulkification consistent with Apex Gov Limits
#450 sendBetterEmail - Update Examples to reflect CPE Labels and Functionality
#449 Mass Email Bug


ericrsmith35:
Jack,

#3 - I struggled with this as well.  I gave up and created a 2nd attribute
to store the opposite and defaulted that to False.

Eric

alexed1 (approved):

PR #544, opened 2020-12-23
Updates Generate Collection Report to use Field Label instead of API Name

jmdohn:
As a consumer of the emails, it would be best to display the field label so that I can understand what fields correspond to the record in Salesforce.

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,46 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "*{!ComponentName}* - _({!ElementName})_ - {!BugTitle}"
+labels: triage
+assignees: ''
+
+---
+
+<!--
+Thanks for reporting issues to the LightningFlowComponents team!
+Please make sure to fill all the details below so we can process feedback properly.
+-->
+
+### How to use GitHub
+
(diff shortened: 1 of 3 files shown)


danifin:
@jmdohn it's a great addition, can you add an option to display lookup field names instead of the Id?

alexed1:
Hey, thanks for this enhancement! Let me make sure I understand it. The current version creates header rows that display API Names and you've changed that to extract the Labels instead. It seems reasonable that the people consuming reports are more interested in labels. Does that capture this change?

jmdohn:
Yes, that's exactly right.

jmdohn:
This may be better to pull out the specific class to reduce bloat in this class and include in the following:
https://github.com/alexed1/LightningFlowComponents/blob/master/flow_process_components/FlowBaseComponents/force-app/main/default/classes/fbc_FieldDescriptor.cls
https://github.com/alexed1/LightningFlowComponents/blob/master/flow_process_components/FlowBaseComponents/force-app/main/default/classes/fbc_GetFieldInformation.cls

I could abstract it to get relationship information much like what @danifin suggests. What do you think @alexed1?

alexed1:
@jmdohn  Hi Jeremiah. I'm interested but not fully clear on what you're proposing in your comment from December 24. Which functionality do you suggest get pulled out of which class and where do you see it going? Very open to this kind of refactoring though.

jmdohn:
Creating a new PR when changes are made.

PR #546, opened 2020-12-26
Fix Field Name

ericrsmith35:
Changed Field Name from Subject to Id in the force:recordData statement - Not all objects have a field named Subject


ericrsmith35:
@alexed1 If you create a new install package, I'll update the documentation on USF

alexed1 (approved):

PR #550, opened 2020-12-28
Ignore Account Duplicate Rules

solo-1234:
Modifications so that tests will pass when Account duplicate rules are active in the org.


alexed1 (approved):
Thanks for the enhancement!

PR #553, opened 2020-12-29
Updates generate collection report to show field label and display relationship fields

jmdohn:
This update will now display labels to end users in the report component.  It will also allow for usage of relationship fields in the component.  The base components have been updated to pass in only one parameter, the field describe and the class will do the heavy lifting.  This is important so if the class were to be expanded to include additional describe information (like help text), it only requires a minor adjustment versus updating input parameters.

Account Parent relationship:
accountRelationship

Case account relationship:
caseRelationship


flow_action_components/CollectionProcessors/force-app/main/default/classes/GenerateCollectionReport.cls
```diff
@@ -29,14 +29,21 @@ global with sharing class GenerateCollectionReport {
             if (inputCollection != null && !inputCollection.isEmpty() && shownFields != null) {
                 reportString += 'Collection Type: ' + inputCollection[0].getSObjectType().getDescribe().getName() + '\n\n';
                 List<String> shownFieldsArray = shownFields.replaceAll('[^a-zA-Z0-9\\,\\_\\.]', '').split(',');
+                for(String field : shownFieldsArray){
+                    system.debug('Shown Field: ' + field);
+                }
```
jmdohn, line 34: Can we get rid of these? They were for just debugging.  It won't let the PR requestor remove code. @alexed1

adwhite75:
Awesome additions @jmdohn ! Would you mind throwing in some comments for each of the methods in here?

adwhite75:
You may also want to check the reference to fbc_FieldDescriptor - the new managed package version uses a different namespace (new managed version found here: https://unofficialsf.com/introducing-flow-basepacks/)

ericrsmith35:
In the managed package it is just FieldDescriptor

jmdohn:
Added comments, removed a few debugging statements.

This would require the managed package to be updated for this new code to be incorporated.  It would mean that there is a hard dependency on the managed package for this to work.  Trying to reuse code / minimize bloat here, but is that dependency something that is wanted?  I can make it similar to PR 544 where I make classes and handle it in the component itself.

flow_action_components/CollectionProcessors/force-app/main/default/classes/GenerateCollectionReport.cls
alexed1: Hmm. What happens when you try removing them in your local project. Then add and commit? Removing is no different from adding in fit world...

jmdohn: I got it figured out. :-)

alexed1:
Our current philosophy, @jmdohn, is to invest in two key packages, one for actions and one for screen components, and encourage everyone to make heavy use of them as dependencies. To the extent that works, it means that you don't need to bloat your own package. In theory, many users will already have installed these to get some other Flow package to work, so it won't become a big burden even if you're only using them for a single call. So I think that the dependency makes sense here even though in isolation it would make sense for you to include the code.

alexed1:
Pertinent to @ericrsmith35 's comment, if you reference the  FieldDescriptor in the  current managed Flow Actions Base Pack, you need to prepend the namespace like this usf.FieldDescriptor or this usf__FieldDescriptor.

alexed1:
As part of this improvement, @jmdohn, can you draft a blog post on unofficialsf.com that explains the changes and points back to the existing Generate Collection Report at https://unofficialsf.com/convert-record-data-into-tables-for-email-automation-with-generate-collection-report/. Also, please edit the Generate Collection Report post to incorporate your changes. Point me at the drafts when you're done. This does not have to be extensive. Just looking for the key information....

jmdohn:
Updated to include usf namespaceprefix.  This would require a new version of the managed package to be released with the new code for fielddescriptor / getfieldinformation.

Will work on a draft later today.

ericrsmith35:
I'm curious what changes are being made to fieldDescriptor and
getFieldinformation.

alexed1:
I'm happy to build a new Actions Base Pack to reflect these changes to getFieldInformation. Eric, it looks like Jeremiah has generalized the action to pass around an entire describe structure rather than a small number of specific fields

jmdohn:
Did some additional testing from a custom object and the labels are now what you'd expect.  Doing some more modifications to get this displaying correctly and will push in a bit.

ericrsmith35:
I just want to confirm that this won't "break" any existing components that
use FieldDescriptor and GetFieldInformation

jmdohn:
New fixes are in for fixing the headers to now display correctly regardless of field / number of parent relationships:

accountFlowHeaderFix

@ericrsmith35, the changes that were made:
1. Update the fbc_FieldDescriptor to have an input of field schema so that the attributes are determined dynamically, including gathering field related to values
2. Updates the fbc_GetFieldInformation to have a net new method for getting schema describe information and now uses the new constructor mentioned in #1

This _could_ be a breaking change if the legacy constructor is used by custom apex development.  In order to make it not a breaking change, we can retain the legacy constructor.  The only component I found that it was used as an attribute is the Dual Listbox component.
[https://unofficialsf.com/duallistbox/](https://unofficialsf.com/duallistbox/)

alexed1:
good find on duallistbox. Yes, let's definitely reteain the legacy constructor. we can have them both.

jmdohn:
And it's back.

jmdohn:
@alexed1 The post is now drafted and available here - https://unofficialsf.com/wp-admin/post.php?post=21554&action=edit

alexed1 (approved):
I'm going to go ahead and merge this without waiting for the changes to FlowBaseComponents to be pulled out. We don't plan to do any more packages of FlowBaseComponents so it shouldn't be a big deal. If you want to @jmdohn, create a new pull request that makes these changes to FlowActionsBasePack and we'll get them into the next rev of that package.

alexed1:
@jmdohn I ran into errors trying to package this code so I reverted it. 

The first issue is that 2gen managed packaging seems to reject the syntax 'usf__foo' in favor of 'usf.foo'. I changed things to dot notation and it worked.

Then I got this error:
Error  force-app/main/default/classes/GenerateCollectionReport.cls  Method does not exist or incorrect signature: void describeSobjects(String) from the type usf.GetFieldInformation (47:103)

if you look at that line, you see this:
Map<String, usf.FieldDescriptor> originalObjectfieldMap = usf.GetFieldInformation.describeSobjects(String.valueOf(inputCollection[0].getSobjectType()));

however, describeSObjects seems to expect a List of strings but it's only being passed a single string. 

I was unable to push the code to a scratch org because of these errors.

Jeremiah, when you get a chance, can you take a look and create a new pull request that deploys successfully to orgs that have Actions Base Pack 2.11.0-0" or higher?

jmdohn:
You bet.  I can take a look sometime next week and update to use Flow Actions Base Pack.

PR #579, opened 2021-01-27
Compare Records Action

adwhite75:
Just wrapped up putting together a neat action that I call 'FindCommonAndUncommonRecords' - It compares two like/unlike record collections and spits out records that are common and unique between the two. Best part is that you can compare two collections of different sObject types AND specify the Ids/fields to compare against.  Best part is it utilizes zero SOQL.

it takes 4 inputs and provides 4 outputs.

**Inputs:**
sourceRecordCollection (Record Collection)
sourceUniqueID (String)
targetRecordCollection (Record Collection)
targetUniqueID (String)

**Outputs:**
sourceUniqueRecordCollection - The unique records in the source collection when compared against the target collection
sourceCommonRecordCollection - The shared records between the two collections (returns the source sObject type)
targetUniqueRecordCollection - The unique records in the target collection when compared against the source collection
targetCommonRecordCollection - The shared records between the two collections (returns the target sObject type)


adwhite75:
Pausing while I make a quick improvement in constructing the field map

alexed1 (commented):
it looks like you've got two copies here, one for collectionprocessors and one for flow_action_components as a 'standalone'. we generally try to avoid that though because they tend to fall out of sync, with someone coming along and updating one but not the other. So I suggest you delete the standalone one...

adwhite75:
@alexed1  fixed the dupe! Must have pasted into the wrong place twice. Still getting used to Macs :)

alexed1 (approved):
great stuff, Adam!

PR #584, opened 2021-01-30
Updates Generate Collection Report for Relationship Fields + Field Labels

jmdohn:
Updates Generate Collection Report to have relationship fields be displayable, labels for headers and per PR 553, now uses usf actions as opposed to legacy fbc (https://github.com/alexed1/LightningFlowComponents/pull/553).


alexed1 (approved):
very nice....

jmdohn:
> @jmdohn this is really great. Would you be up for making a feature where you can specify the names of each column similar to how we do it for Datatable? Great that we can pull in related fields - I think a great improvement to that would be to specify the name of the field. For example if I pull in Owner.Name, it shows 'Name' as the column header.

I don't see why it wouldn't be possible.  I'd have to dig into the datatable a bit more but I think it'd be a relatively simple addition.

PR #591, opened 2021-02-10
Added missing -meta.xml and associated static resources

sharadmaheshwari:
This should hopefully fix the build issue


alexed1:
Here's package 1.0.0.0 https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G000004PucoQAC

can you try it out and verify that it works as expected?

PR #593, opened 2021-02-10
Updated FlexCardFlow

jdayment:
Updated API versions and removed SRC property from FlexCardFlow xml not needed.


jdayment:
Closing and resubmitting a new one too include Field Label Fix.

PR #612, opened 2021-02-22
Update FlexCard to support more than account object

jdayment:
I added an inputOnly property to provide the object API Name for the js objectInfo wire I couldn't figure out how to get it from the "T" propertytype wasn't sure if that's possible or not.


ericrsmith35:
I've only been able to get the object API name in a LWC by calling out to Apex with a 'record' from the object and having it return the object name.

String objName = record.getSObjectType().getDescribe().getName();

alexed1 (approved):

PR #628, opened 2021-03-07
New CheckValidation Flow Screen Component

ericrsmith35:
https://ericsplayground.wordpress.com/2021/03/07/validation-checker-flow-action/

I'll add this post to USF as soon as the source gets uploaded.


alexed1 (approved):
Impressive.

PR #646, opened 2021-03-15
Iframe Component for Screen Flows

jdayment:
Had someone ask if there were any out there didn't find one so I created one thought I would share. Simply put in your url and dimensions and the website displays in your flow screen gotcha's will be any csp settings in your org.


jdayment:
Will resubmit later.

PR #660, opened 2021-03-24
changes by Hijlko to CSVToRecords. Adds capabilities

alexed1:


alexed1:
This PR replaces https://github.com/alexed1/LightningFlowComponents/pull/626
@hijlko

alexed1:
From PR 626:

I donwloaded the package for converting CSV files to records. I found it very usefull but had to change the apex code to be able to import CSV files of the European region. Standard US csv uses a comma as a field separator while most European csv's use a semicolon as a field seperator. There are more diferences between the regions.
Date: YYYY-MM-DD for US and SalesForce standard but DD-MM-YYYY for European standard
Thousand/decimal separator: In US the point is used as a decimal separator an a comma as a thousands separator.
In Europe a comma is used as a decimal separator and a point as thousands separator.

That's why I made some changes to the code/componet so you're able to set the

field separator
Thousand seperator
Decimal Separator and
Currency symbol
The code can now import muli region CSV files
Numbers/currency like 1,2345.56 (US) and 1.234,56 (EU) are converted to 1234.56
Date like YYYY-MM-DD and DD-MM-YYYY (and other separators like YYYY/MM/DD and DD/MM/YYYY) are always converted to YYYY-MM-DD

I'm very new to APEX coding so maybe it can be coded in a better way but it works for me. Maybe you can check it out.
ConvertAllCSVToRecords.txt

alexed1:
@hijlko, I've created a PR by putting your modified class on a different branch. 

Are the new attributes that you've added optional? I.e. if I don't use them, will it still work the old 'American' way?

alexed1:
Can you add a test class, @hijlko?

Hijlko:
TestClass
[ConvertAllCSVToRecordsTest.txt](https://github.com/alexed1/LightningFlowComponents/files/6196099/ConvertAllCSVToRecordsTest.txt)

Hijlko:
The added attributes are not optional

ericrsmith35:
We should consider creating a different Collection Component for this or redo it so all of the new attributes are optional and won't cause any issues with existing implementations if someone upgrades.

Hijlko:
I've made some changes so all the variables are optional now. I've set the string variables to (curInput.<varName> + original string).left(1) so when the Ibnput is Null it should use the original value (',' for field separator, $ for currency symbol, etc. When not null the value that was set in the component is used.
I've also added a new testclass that should get a 77% result. The code is in the attached files
[ConvertCSVToRecords.txt](https://github.com/alexed1/LightningFlowComponents/files/6238522/ConvertCSVToRecords.txt)
[ConvertCSVToRecordsTest.txt](https://github.com/alexed1/LightningFlowComponents/files/6238523/ConvertCSVToRecordsTest.txt)

alexed1:
Thanks, Hijlko. I'll take a look at this as soon as I can get some free time.

alexed1:
@surajp and @Hijlko : you both have submitted great  and different improvements to CSVToRecords. What I'd like to suggest is that you each code review the other's pull request! I'd like to merge them at the same time and then create a new package

alexed1:
Hi @Hijlko  : check out your work featured here: https://unofficialsf.com/from-suraj-pilai-and-hijlko-csv-file-import-gets-some-great-internationalization-improvements/

We ended up moving your code into another pull request, so closing this one. Thanks again for the great contribution! Please let us know if you improve any other Flow extensions.

Hijlko:
Hi,
I don't think the right version of my code was used. I don't see where the 'default' separators/currency is set when no optional seperators/currency are entered.
Besides that, at the description of the thousand and decimal separator, there should be a choice for comma or point/dot, not comma or semicolon.
Here is the code of my last version with the right settings.
Kind Regards,
Hijlko Roosjen.


________________________________

PR #688, opened 2021-04-14
Update to flowCarousel

jdayment:
Updates flowCarousel to allow for a hyperlink to be added to the slides also fixes some styling parameters. (I am not sure how to remove a file from a PR but there is one in here for fileUploadImproved as well it is a just an additional space that was driving me nuts when looking at the code doesn't impact performance or functionality it was just cosmetic. )


alexed1:
since you're confident that the file uploaded change is not problematic, we can just leave it in and it will sit quietly in the code base until the next time we repackage that project.

jdayment:
100% confident

alexed1 (approved):

alexed1:
flowCarousel 1.1. https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G0000047xF7QAI

PR #703, opened 2021-04-24
datatable picklist highlight fix

JerrycsPoon:
Updated the behaviour of datatable picklist highlight when cell is edited. This will now resemble standard behaviour, where the box is highlighted but the dropdown isn't. The box highlight is controlled by salesforce by manipulating the fieldValues of the datatable.


JerrycsPoon:
hi @ericrsmith35 , after this change, the picklist highlight behaviour should resemble standard salesforce. The highlight will also correct clear on save.

PR #708, opened 2021-04-30
Workflow Buttonset (take 2)

fromsteinsfdc:


alexed1:
This is a great first effort, David, and done very quickly considering how many technologies you have to grapple with. You did make one big error, though. You created your flowButtons project at the root of the flowcomponents repo instead of in flow_screen_components/flowButtonBar. If I merge this in as is, you would end up getting created here: 

image


to fix this:
1) close/throw away this pull request. pull requests are cheap. don't bother trying to fix it.
2) in your local clone of the flowComponets repo, create a new folder in flow_screen_components to hold your project. because there's already a 'flowButton', I suggest you use flowButtonBar for a little more distinctiveness. 
3) move your project into the new flowButtonBar directory
4) do "git add ." from the flowButtonBar directory to add all the files and then commit and push them, and create a new PR

PR #709, opened 2021-04-30
flowIframe

jdayment:
I have had a couple people ask if I knew of a solution so I made one. Real simple LWC that iframes a URL in flow.


alexed1 (commented):
this looks great, Josh. Can you draft up a post? I would focus it on the benefits (why would you want to use an iframe? What can you do with one that you can't do without one?). Too many iframe discussions assume that the reader fully understands them, and that's usually not the case.

jdayment:
Sure thing!

alexed1 (approved):

PR #715, opened 2021-05-05
Dt picklist

ericrsmith35:
Current source code for datatable v3.2.0


alexed1 (commented):
renaming files for existing extensions makes me very nervous these days...we've just had so many issues with upgrading failing. What's the motivation for renaming these classes? Do we know that upgrading will work without problem?

ericrsmith35:
The base component (datatable) was not renamed.  The only files renamed were ones referenced by datatable.  The reson for this is to reduce the conflicts users are starting to see when insalling multiple unlocked packages where different developers have chosen the same names for various pieces of their components.

ericrsmith35:
There is another developer out there who has a datatable component geared towards use by developers and it can't be installed in the same org as mine because he also uses "datatable".

ericrsmith35:
BTW, I have installed this update, as have others, and I've seen no issues with the renaming.

PR #718, opened 2021-05-06
output default recordtype to the flow

Tamarer:
@alexed1  a small but useful addition to this action to return the default record type to the flow


alexed1 (approved):
Nice enhancement, Tamar. Great to see you active again in this project!

Tamarer:
Thank you @alexed1  can you please update the installable package with this update, I don’t have access, thanks!

PR #736, opened 2021-05-21
Merging iconSelector into fsc_pickIcon

fromsteinsfdc:


ericrsmith35:
@fromsteinsfdc How will your changes affect existing implementations that are using fsc_pickIcon?  If someone upgrades to the new version and makes no changes to existing components using fsc_pickIcon, will they still function the same way?

fromsteinsfdc:
@ericrsmith35 that's a really smart (and in hindsight obvious) consideration I completely ignored. There would indeed be some problems:

- I replaced the `accordionMode` boolean property with a string `mode`, and since I did not specify a default value for it, it will show an error message. Possible fix: I add `accordionMode` back with a setter that sets the `mode` value to `accordion`. (I'm not sure if I'd have to add it back to the XML file, or if simply adding back the @api property would work.)
- I changed `hideActionIcons` to `excludeActionIcons` to match the other `exclude____Icons` from my original component, so anyone who had set `hideActionIcons` to true would lose that setting. Possible fixes: I can just change `exclude` to `hide` for all 4 properties, or we can set `excludeActionItems` to default to true, 'cause IMO they're not wanted most of the time.

I don't think there would be any other issues: I adapted my code to use your `iconName` property and kept your same selection event.

ericrsmith35:
Unfortunately, once one starts adding enhancements to an
existing component, the most efficient way of handling attributes isn't
always possible.  I think that once an attribute is in place, new versions
need to account for it for backwards compatibility.  I like where you're
going with the new attributes and I think all that needs to be done is to
accept and convert the older ones to the corresponding new values and
document that they are obsolete and should not be used in the future.

Is there a noticeable improvement in speed with the new combobox option
over the accordion option?  Will the combobox option display ok in a narrow
CPE canvas?

Thanks,
Eric Smith

alexed1:
This is a good learning experience, @fromsteinsfdc. I should have brought up the issue of backwards compatibility when I asked you to add your new functionality to the existing component. In general, your new plan sounds good. I think you may run into problems if you try to set the default of excludeActionItems to true because of lwc limitations, and think maintaining naming consistency with the original 'hide***' is best here.

You definitely need to keep all the original public attributes like accordionMode, keeping them exposed in the meta file. There's no elegant way to handle trying to go from a boolean like accordionMode to a multiple value solution, and for that reason, it's a best practice to try and avoid booleans in general for public configuration whenever you're dealing with anything like a mode. Anytime I find myself thinking 'i have two modes here' I always immediately run to a string value because too many times I've thought there would be only the two and later wanted to add a third. In this case, try doing what you propose. In the docs we can encourage people to use the new displayMode attribute and stay away from accordionMode, but note that accordionMode will continue to work by setting displayMode.

alexed1:
I'm going to close this pull request....just open a new one when you're ready.

PR #745, opened 2021-05-27
Now shows a pen icon before picklist is editable

JerrycsPoon:
- Mimics the behaviour of other editable cells, where a pen icon is displayed when hovered
- when the pen icon is clicked, only then the picklist is editable
- selecting a new picklist value will revert the cell back to view mode
- clicking away from the picklist will revert the cell back to view mode


JerrycsPoon:
hi @ericrsmith35, I've took your suggestion and made the picklist cell behave similarly to other cells. Hovering the cell will now allow the user to click on the pen icon, which switches the cell to edit mode and display the picklist for edit.
I've tried my best and implemented so that it behaves the same as other editable cells

alexed1:
@ericrsmith35 pls review.

ericrsmith35:
I will be reviewing this weekend.  Thanks Jerry.

ericrsmith35:
I have added this code to the v3.2.2 branch.  https://github.com/alexed1/LightningFlowComponents/tree/v3.2.2

PR #773, opened 2021-06-15
Create USF_Flow_Action_Collection_Actions.permissionset-meta.xml

ericrsmith35:
Permission Set that includes all of the Apex Classes in the Collection Actions group.  Please create a new package for Collection Actions so this Permission Set will be included with the install.

This will make it easier for Admins to provide access to Users running Flows that use any of these actions.


alexed1 (approved):

alexed1:
1.29 Collection Actions https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G0000047xKlQAI

PR #779, opened 2021-06-18
Quick Lookup

ericrsmith35:
I have refactored the lookupFSC component to a new version call Quick Lookup. The new version is built using a Generation 2 unmanaged package and should no longer have any upgrade issues. You can install and use the new version without having to delete or change any currently active flows.


alexed1:
Is the plan that all packages of this will be built on the same lexhost org I use for other unmanaged packages?  that seems like a good idea.

ericrsmith35:
I think that is a great plan.  This was built on lexhost.  I've created a
few simple batch files for myself to make it easier to create and update
packages on that site.  See the following files in the datatable project
folder
<https://github.com/alexed1/LightningFlowComponents/tree/master/flow_screen_components/datatable>
.

   - SetVersion.bat
   - ShowVersion.bat
   - CreateNewVersion.bat
   - Promote.bat
   - Details.bat

I should also create one for NewPackage (sfdx force:package:create -v
lexhost --name "PackageName" --path force-app --packagetype Unlocked) and
update these to allow the PackageName to also be a variable attribute.

Eric

alexed1 (approved):

alexed1:
ok. sounds great.
 I'll leave it to you to package and update.

PR #784, opened 2021-06-22
Updated GetPicklistValues to be batchable

Garywoo:
The GetPicklistValues action was erroring when performing actions during a batch update. This change fixes the issue.

If you performed a mass update, and a flow containing this action was triggered on record update, you would get the following error message:

> The number of results does not match the number of interviews that were executed in a single bulk execution request.

The action has been updated to allow it to run correctly in such scenarios, by looping though the requests, instead of only executing on the first.

The API version of the class has also been updated to v52.0.

This issue was first raised in the comments of the flow action: https://unofficialsf.com/flow-action-get-picklist-values/#comment-5123649046


alexed1 (approved):
Thanks for fixing this, Gary!

PR #787, opened 2021-06-22
Temporary fix for #786 - cached TemplateName Overflows

### LINKED ISSUE #786: SendBetterEmail - BetterSendEmailUtil- Too Many SOQL Queries error from "cache" email templates
@Luke Freeland is having an issue with Send Better Email.  Using the BetterSendEmail apex action from a scheduled flow with cases to send reminder emails via email templates. This is my first instance of using scheduled flow with apex actions to try and send emails. It looks like we're hitting apex limits because all the flow interviews are running together instead of each record running the flow interview sequentially. We're hitting a Too Many SOQL Queries error and we think it's because the BetterSendEmail apex class is calling the BetterSendEmailUtil.getMapTemplateNameByEmailTemplates function to "cache" the email templates. However, we're using email template ids so there's no need to query the email templates so the email template names is empty but is still issuing the query which will return no email templates. We think since Apex Actions are called once per interview, it's calling this multiple times.
The fix I'm trying to apply is this:

`    public static Map<String, List<EmailTemplate>> getMapTemplateNameByEmailTemplates(
        Set<String> templateNames,
        Map<Id, EmailTemplate> mapIdByEmailTemplate
    ) {
        Map<String, List<EmailTemplate>> mapTemplateNameByEmailTemplates = new Map<String, List<EmailTemplate>>();
        if (templateNames == null ||
            templateNames.isEmpty()) {
            return mapTemplateNameByEmailTemplates;
        }`

jdpond:
May Fix  #786
Possible solution as identified for scheduled flow caching issue as recommended by 
Luke Freeland

### DIFF
--- a/flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmailUtil.cls
+++ b/flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmailUtil.cls
@@ -2,13 +2,14 @@
  * @description       : SendBetterEmailUtil - helper utilities for SendBetterEmail
  * @author            : Refactored from Alex Edelstein
  * @group             : 
- * @last modified on  : 02-17-2021
+ * @last modified on  : 2021-06-22
  * @last modified by  : Jack D. Pond
  * Modifications Log 
  * Ver		Date		Author				Modification
  * 2.00.02	10-06-2020	Mohith Shrivastava  Refactored from SendBetterEmail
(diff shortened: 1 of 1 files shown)


alexed1 (approved):

alexed1:
I've merged this and geenrated a new 2.1.4 package:  https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G0000047xLUQAY

do you want to publish it immediately? @jdpond

PR #795, opened 2021-06-29
Make Convert CSV To Records Action More Spec Compliant

surajp:
Made changes to have the action support more CSV files according to spec. The following scenarios are now handled:
- Rows/File beginning and ending with quotes
- Newline characters within columns
- Null columns at the end of a row

### DIFF
--- a/flow_action_components/ConvertCSVToRecords/force-app/main/default/classes/ConvertCSVToRecords.cls
+++ b/flow_action_components/ConvertCSVToRecords/force-app/main/default/classes/ConvertCSVToRecords.cls
@@ -17,9 +17,9 @@ public with sharing class ConvertCSVToRecords {
 
                 Blob csvBlobContent = getCsvContent(curContentDocumentId);
                 String csvStringContent = blobToString(csvBlobContent, 'ISO-8859-1');
-                String tempRepl = '~`~', newlineRepl = '-`-';
+                String commaRepl = '~`~', newlineRepl = '-`-';
                 csvStringContent = csvStringContent.replaceAll('(\r\n|\r)', '\n');
-                csvStringContent = removeCommas(csvStringContent, 0, tempRepl, newlineRepl);
+                csvStringContent = removeCommas(csvStringContent, 0, commaRepl, newlineRepl);
                 String[] csvRows = csvStringContent.split('\n');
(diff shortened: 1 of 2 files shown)


surajp:
Apologies for the formatting changes, wasn't sure what your prettier settings are. If you can let me know I can resubmit with the right settings. For the tests, I didn't add any new test methods, and changed the `createCSVBlob` method to create csv records that would cause these changes to be tested

adwhite75:
May not be needed for this PR, but I think a valuable flag / input would be the option to trim all of the fields of any leadin/trailing spaces. This could help in scenarios where a column might have a unique identifier but the user accidentally had an extra space in it, which would cause downstream issues.

surajp:
@adwhite75 that's a good idea. I added an optional `isTrim` input attribute to allow users to decide if they would like leading and trailing spaces removed from field values. Additionally, field names in the header are always trimmed since we never have spaces in field names

surajp:
Have added positive and negative tests for trimming. Also, forgot to mention another thing, this PR also includes changes to handle double quotes within the column, which according to CSV spec, is represented by consecutive double quotes (`""`) within a column enclosed in double quotes

adwhite75:
This is great, thanks!

alexed1:
@surajp thanks so much for these improvements. Can you create a google doc that lists and describes the functional changes you've made, for admins consuming the action, and also the new parameter that has been added? You don't have to discuss the tests. I'll use that doc in a blog post that describes your changes. Which spec are you referring to as 'the CSV spec'? 

Really appreciate your work to improve this extension!

alexed1:
@surajp and @Hijlko : you both have submitted great  and different improvements to CSVToRecords. What I'd like to suggest is that you each code review the other's pull request! I'd like to merge them at the same time and then create a new package

alexed1:
Hi @surajp : check out your work featured here: https://unofficialsf.com/from-suraj-pilai-and-hijlko-csv-file-import-gets-some-great-internationalization-improvements/

We ended up moving your code into another pull request, so closing this one. Thanks again for the great contribution! Please let us know if you improve any other Flow extensions.

surajp:
Thanks for the update and the shout out @alexed1

PR #812, opened 2021-07-12
FlexCard updates

jdayment:
Added Label and cleaned up XML properties in prep for CPE, made objectAPIName required.


jdayment:
Closing and resubmit ing just noticed an errant file

Issue #829, opened 2021-07-28
sendBetterEmail - Attachments - Sends all versions as attachment - should only send latest Version


jdpond:
### Steps to reproduce

1. Add a ContentDocumentId to the CDL Collection for a document with multiple versions to a flow containing a sendBetterEmail Action
2. run the flow

### Expected behaviour
Each message should have the most current version of the prescribed attachment 

### Actual behaviour
"it's sending out the same attachment 23 times together with my email"

### Screenshots
N/A

### Debug Logs
[Log Snap Shot](https://user-images.githubusercontent.com/271131/127361142-3e7f0df9-6000-493f-a9b5-c2ff37fb04d7.png)

jdpond:
SendBetterEmailUtil.getMapContentDocumentIdByContentVersions selects documents to be added with the following query:
```
	SELECT Id, PathOnClient, Title, VersionData, FileType, ContentDocumentId, FileExtension
	FROM ContentVersion
	WHERE ContentDocumentId IN :contentDocumentIds
```
which will select all versions of all documents with specified Ids.

Will add VersionNumber to the query and then parse out all but the latest version number out before returning the map.

jdpond:
Temporary workaround is to make sure that all specified documents have only one version - copy the latest version into a new document then use that documentId in the CDL

PR #830, opened 2021-07-29
Temporary fix for #829 and #786 

jdpond:
Temporary fix for #829 - Sends all versions as attachment - should only send latest Version
Temporary fix for #786 - cached TemplateName Overflows


alexed1:
Thanks for these improvements, Jack!

alexed1 (approved):

PR #833, opened 2021-07-30
CCI and Snowfakery Setup

jmdohn:
Removed failing deployment components in collection processors and RefreshCollection Action
Added cumulusci yaml
Added option for running tests

### DIFF
--- a/flow_screen_components/FlowScreenComponentsBasePack/force-app/main/default/lwc/fsc_actionList/fsc_actionList.html
+++ b/flow_screen_components/FlowScreenComponentsBasePack/force-app/main/default/lwc/fsc_actionList/fsc_actionList.html
@@ -23,23 +23,25 @@
                 </div>
             </div>
         </template>
+
         <template if:true={displayMenu}>
-         <lightning-button-menu label={buttonLabel}
-         alternative-text="Actions"
-         tooltip="List of available actions for this specific record click on the name of the action to launch"
-         onselect={launchFlow}>
-    
-            <template for:each={flowData} for:item="flow">
-                <lightning-menu-item
-                    id={flow.apiName}
-                    label={flow.label}
-                    value={flow.apiName}
-                    key={flow.label}>
-                </lightning-menu-item>
-            </template>
-    
-            </lightning-button-menu>
-        </template>
+            <lightning-button-menu label={buttonLabel}
+            alternative-text="Actions"
+            tooltip="List of available actions for this specific record click on the name of the action to launch"
+            onselect={launchFlowMenu}>
+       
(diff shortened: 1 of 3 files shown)


alexed1:
Jeremiah, I'm thinking that, to keep the main project README from getting too technical, we take your cumulus script docs and put them in a different readme called something like deploymentTools.md, and then reference that from the main Readme. What do you think?

jmdohn:
@alexed1, it may make sense to include in a contrib.md which is a standard practice as the intent is to make the process easier for PR and development?  Here's a sample [contrib.md](https://gist.github.com/PurpleBooth/b24679402957c63ec426).  What do you think?

alexed1:
Sure great idea. Tuning of the repo's main pages is overdue

jmdohn:
I added the contrib.md and updated the readme to direct to it.

contrib.md
```diff
@@ -0,0 +1,46 @@
+# Contributing
+
+When contributing to this repository, please create a pull request.  If you would like to have interest
+of an idea guaged, please create an issue or request feedback in the flow projects slack workspace.
+
+## Development
+
+For making it easier to create pull requests, you can use the below commands to install the components in scratch orgs for development purposes.
+
+You must install [cumulus ci](https://cumulusci.readthedocs.io/en/stable/intro.html) to run the below commands. You can set up a scratch org with the components in this repository by running the below command:
```
alexed1, line 10: Thanks @jmdohn. I fear I may not have understood your thought process. Here's the issue: this text can be thought of as the path that a new prospective contributor might follow. It's a steep path and I want to encourage new contributors. The problem with inserting cumulus ci stuff here is that most new contributors won't have experience with ci/cd or cumulus, so they'll have to puzzle through this to figure out that it's optional.  So what I'd like to do is decouple the cumulus stuff into a separate document so that the path of new contributor just sees something like "if you are using cumulus ci, click here for some useful utilities". Thoughts?

jmdohn: I'm hoping to automate many of the PR requested steps with github actions so they don't need to remember / do the steps below for adding a "net new or archiving an existing" component and auto adding it to the appropriate action group based on directory for cci. Although, this will take at least a few weeks to complete depending on my bandwidth. I definitely appreciate the need to keep it low level of entry to folks who may not be familiar with cci and am totally okay with breaking it out into a separate document and referencing that in the readme.

Happy to also add details in that same document for those using just sfdx for their work.

Once I get some of the more fancy CCI automation done, I'm happy to take some time to run you through it before creating that PR because it's some pretty geeky stuff that most contributors won't need to interact with in nearly every circumstance (unless they really really want to).

PR #838, opened 2021-08-06
datatable: don't show radio button in data table when tableData.length is 1

steot:
When the table only contains a single row (tableData.length === 1), maxRowSelection is set to 1 and the row checkbox becomes a radio button, which cannot be deselected by the user.

This should not happen and radio buttons should only replace checkboxes when the option "Single Row Selection (Radio Buttons)?" (singleRowSelection) is activated explicitly.


alexed1:
Thanks for the contribution, Stefan! @ericrsmith35 can you take a look?

ericrsmith35:
Thanks for the update.  This has been addressed in the current version (V3) of Datatable.

steot:
@ericrsmith35
Thanks for the info, however I can't find this in the current version (V3) of Datatable. The code there is exactly the same as in the previous versions and will lead to the same behavior I mentioned. Or am I confusing this somehow?
https://github.com/alexed1/LightningFlowComponents/blob/87547c07f9f9ca4c63e912e03d99252fdb01dcc7/flow_screen_components/datatable/force-app/main/default/lwc/datatable/datatable.js#L600

ericrsmith35:
The Radio Button for selection when there is only a single row in the
datatable is built-in functionality of the base Salesforce datatable
component.

Version 3 of the code adds an optional Clear Selection button to address
this issue.  See line 1397.
LightningFlowComponents/datatable.js at
87547c07f9f9ca4c63e912e03d99252fdb01dcc7 · alexed1/LightningFlowComponents
(github.com)
<https://github.com/alexed1/LightningFlowComponents/blob/87547c07f9f9ca4c63e912e03d99252fdb01dcc7/flow_screen_components/datatable/force-app/main/default/lwc/datatable/datatable.js#L1397>

ericrsmith35:
Good catch.  I'll update the behavior in the next release.

ericrsmith35:
This has been implemented in the latest release of Datatable (v3.3.2)
https://unofficialsf.com/datatable-lightning-web-component-for-flow-screens-2/

Issue #837, opened 2021-08-02
QuickChoice - selectedLabel output attribute - Does not get set when a default value is set and accepted


Garywoo:
When passing in a default `Value` attribute into the QuickChoice component, if the user accepts the default value, and does not select a different value, the output attribute `selectedLabel` returns "null" instead of the expected default values' label.

### Steps to reproduce
1. Create a new screen flow.
2. Create two new string collection variables to be used for QuickChoice labels and values.
3. Add at least 2 assignment elements, adding different string values to each of the QuickChoice label and value string collection variables.
3. Add a screen element containing a QuickChoice component to the flow, set the display type to Picklist, and select the datasource "Two String Collections (Labels and Values)".
2. Set the QuickChoice value to equal the first value added to the QuickChoice values string collection variable.
3. Add an assignment element after the screen element to help with debugging.
4. Set a new variable to be equal to the `selectedLabel` attribute from the QuickChoice component. 
5. Debug the flow.
6. Do not choose a value in the QuickChoice picklist, just leave the default option selected.
7. Click Next.
8. Observe the value of `selectedLabel`.
9. Click Run Again.
10. This time, actually select the 1st option in the QuickChoice picklist, even though it's already the default value.
11. Click Next.
12. Observe the value of `selectedLabel`.
13. Click Run Again.
10. This time, select the 2nd option in the QuickChoice picklist.
11. Click Next.
12. Observe the value of `selectedLabel`.

### Expected behaviour
The `selectedLabel` attribute should always be set to the label of the value of the QuickChoice picklist, even when left as the default.

### Actual behaviour
The `selectedLabel` attribute is only set to the label of the selected QuickChoice picklist value when a non-default value is selected. If the value is left selected, then the `selectedLabel` returns "null".

### Screenshots
Here's a screen recording of the reproduction steps above being followed.

![QuickChoice selectedLabel issue](https://user-images.githubusercontent.com/10392194/127895129-ea03c400-cd26-4586-a765-82e0b5c75bc1.gif)

ericrsmith35:
The Selected Value should be used for output from Picklists and Radio Buttons.  Selected Label is used for the output from Visual Text Boxes.

ericrsmith35:
This will be fixed in the next release of the FlowScreenComponentsBasePack.  Pull Request #844

Garywoo:
Thank you eric!

PR #844, opened 2021-08-06
Fix Selected Label Bug #837

ericrsmith35:


alexed1 (approved):

alexed1:
as always, nag me if I don't get around to reving the package in a week or so...

ericrsmith35:
Will do.

PR #845, opened 2021-08-06
Updated barcodeScanner

jdayment:
General clean up and adds additional functionalities. 

"label" Add a custom label before the scannedBarcode in the main screen
"autoNavigate" When marked as true flow will automatically move to next action/screen in your flow upon successful scan


jdayment:
I'll add this to my flexCard CPE PR

PR #853, opened 2021-08-09
Adding static choices as a datasource for quickchoiceCPE

fromsteinsfdc:
Also some small changes to FlowButtonBar, including adding support for a default value.


alexed1 (approved):
looks like a great PR. I'll pull it local and push to a scratch org to see if there are any problems

PR #862, opened 2021-08-13
sendBetterEmail 2.1.7 Planned Release, Fix #855

### LINKED ISSUE #855: SendBetterEmail - Allow Deployment if Enhanced Email Disabled
**Is your feature request related to a problem? Please describe.**
SendBetterEmail v2.0.4 introduced a dependency on Enhanced Email,  in test method t031_canSendEmailwithInReplyTo. This dependency prevents installation in an org with Enhanced Email disabled since the EmailMessage object is not available in that scenario. The EmailMessage object is used only in the test method, and nowhere else in the package.

**Describe the solution you'd like**
The test validates that a messageIdentifier can be put in the SendBetterEmailRequest().InReplyTo field, but does not confirm that the resulting message is actually threaded to the original method. The messageIdentifier is currently created by sending a SingleEmailMessage to create a valid messageIdentifier. Since the only [requirement for such an identifier](https://datatracker.ietf.org/doc/html/rfc2822#section-3.6.4) is that it be unique and correctly formatted, it should be sufficient for test purposes to create a string of 
`'<' + String.valueof(DateTime.now().getTime()) + '@foo.com.invalid>'`

**Describe alternatives you've considered**
I can't see any alternatives since the package does not install with that reference unless Setup > Email > Enhanced Email is enabled.

**Additional context**
Here is the error email on the failed install:
```
Your request to install package "SendBetterEmail ver 2.0.6" was unsuccessful. None of the data or setup information in your salesforce.com organization was affected. 

If your install continues to fail, contact Salesforce CRM Support through your normal channels and provide the following information. 

Organization: XXX (XXX) 
User: XXX (XXX) 
Package: SendBetterEmail (04t5G000003rUZT) 
Error Number: 179535045-29977 (-810699787) 
Problem: 

1. Invalid type: EmailMessage 
SendBetterEmailTest: Invalid type: EmailMessage 

2. Variable does not exist: MessageIdentifier 
SendBetterEmailTest: Variable does not exist: MessageIdentifier 
```

### LINKED ISSUE #873: sendBetterEmail - Add New Action assign to Email Category and version updates
Add sendBetterEmail to the Email category for creating New Action in Flow Builder instead of Uncategorized
Update API to 52.0
Change Version from 2.0.6 to 2.1.7
Package dependencies updated:
  FlowActionsBasePack@2.32.0-0
  FlowScreenComponentsBasePack@2.5.5-0
image

jdpond:
Next Release (2.1.7):

Fix #855 Allow Deployment when Enhanced Email Disabled
Fix #873 Add New Action assign to Email Category and version updates


alexed1:
how is this going? looks like some great fixes....

jdpond:
Let's have a discussion on cadence for releases.  These are done and ready to go, but how often do we want to do releases?

alexed1 (approved):

PR #865, opened 2021-08-13
3.3.2

ericrsmith35:
Datatable v3.2.2 Source Changes

**Updates:** 
-   Converted interface elements to Custom Labels so they can be Translated 
-   Added support for Screen Readers for the visually impaired
-   Updated all component API versions to 52.0

**Bug Fixes:** 
-   Added missing variable in the CPE to support automaticOutputVariables
-   Check for CurrencyConversion returning null values before committing any changes
-   Fix filtering on Date and Datetime column types
-   Keep Checkbox selection instead of Radio Button when table size is 1 and Single Row Selection is not activated


alexed1:
fyi, in Sprint '22 look for the ability to translate select attribute values in lwc's via Translation Workbench. The developer will enable specific attributes by adding 'translatable=true' to the lwc meta.

Issue #861, opened 2021-08-12
FlowScreenComponentsBasePack - Flow Combobox - Error When Selecting a Value from a Record Variable


Garywoo:
While setting values in a Flow Comobobx component, within the CPE of an action or screen component, if you try and use a value from a record variable (i.e. VarOppoprtunity.Id), the following error occurs:

> Something went wrong
> Save your flow and refresh the page. If the problem persists, contact Salesforce Customer Support. Error ID: -901016616.
> 
> If you're working on a screen component, check your component configuration for invalid values.
>
> `This page has an error. You might just need to refresh it. [LWC component's @wire target property or method threw an error during value provisioning. Original error: [Cannot read property 'toLowerCase' of null]] Failing descriptor: {markup://c:fsc_flowCombobox}`

This occurs both when adding a new actions featuring a Flow Combobox in its CPE to the canvas, and when editing existing actions with a Flow Combobox. If a value like `{!VarOppoprtunity.Id}` exists in the Flow Combobx field already, then the error is immediately shown open opening the action editor.

The issue can be worked around if you paste the full path to the field you want in the variable, into the Flow Combobox field, e.g. `{!VarOpportunity.Name}`. Typing it manually causes the error to occur upon every keypress after the variable name, so pasting the path avoids this.

This issue was originally reported in issue #856, however this is not specific to the SendBetterEmail action, and is common to all actions and screen components using the fsc_flowCombobox component.

### Steps to reproduce

1. In a new screen flow, add a new record variable for any object, i.e. opportunity.
2. Add a new screen element to the canvas, and open it.
3. Add any custom component to the screen that uses a CPE, which features a Flow Comobox component, i.e. QuickChoice.
4. Click into the Flow Combobox field, i.e. _Value (Default or Existing)_, and click the **>** next to the record variable name.

### Expected behaviour

*   The record variable expands, and you can select a merge field from the object.

### Actual behaviour

*   The reported error occurs.

### Environment

I have reproduced this error on the following package versions:

- FlowScreenComponentsBasePack
  - 2.5.1
  - 2.5.3
- FlowActionsBasePack
  - 2.29
  - 2.31

### Screenshots

Here's the error occurring, as in the reproduction steps above:

![2021-08-12_22-26-05](https://user-images.githubusercontent.com/10392194/129272410-81d4f71b-d78b-444f-9ead-ab36465d15b1.gif)

jdpond:
[...]

Issue #856, opened 2021-08-10
SendBetterEmail - CPE - Error When Selecting a Value from a Record Variable


Garywoo:
While setting values in the CPE fields within the action, if you try and use a value from a record variable (i.e. `VarOppoprtunity.Id`), the following error occurs:

> Something went wrong  
> Save your flow and refresh the page. If the problem persists, contact Salesforce Customer Support. Error ID: -901016616.
> 
> If you're working on a screen component, check your component configuration for invalid values.

```
This page has an error. You might just need to refresh it. [LWC component's @wire target property or method threw an error during value provisioning. Original error: [Cannot read property 'toLowerCase' of null]] Failing descriptor: {markup://c:fsc_flowCombobox}
```

This occurs both when adding new actions, and when editing existing actions. The error occurs when selecting a record variable in any of the CPE fields.

### Steps to reproduce

1.  In a new flow, add a new record variable for any object, i.e. opportunity
2.  Add the SendBetterEmail action
3.  Click in any field in the CPE, i.e. Sender Display Name, and click the > next to the record variable

### Expected behaviour

*   The record variable expands, and you can select a field from the object.

### Actual behaviour

*   The reported error occurs.

### Screenshots

Here's the error occurring:

![](https://user-images.githubusercontent.com/10392194/128894464-62ee9bb6-7de3-4836-b8e2-6e9329ba8f70.gif)

### Environment

*   SendBetterEmail - 2.1.6
*   FlowActionsBasePack - 2.31
*   FlowScreenComponentsBasePack - 2.5.3

jdpond:
Uggh.  I saw this in testing, then it mysteriously disappeared.  @ericrsmith35 Can you help me look at this one?

jdpond:
I tested this with String, StringCollection, and Object Collection (CDL) in the release test environment for 2.1.6.  Seemed to work.  Could this be something with Opportunity?
[Tested in Scratch Box Video](https://user-images.githubusercontent.com/271131/128907545-5919809c-86f0-451c-bbd3-e3da5da0513d.mp4)

Garywoo:
@jdpond Yes, I upgraded both Flow Actions Base Pack and the Flow Screen Components Base Pack to the latest versions today before upgrading SendBetterEmail. This is when I started noticing the issue occurring.

The issue is occuring for any object in the record variable, not just opportunities. I also tried leads, and accounts.

[...]

PR #870, opened 2021-08-16
Allow for SObject field type to be null (CloneSourceId)

ericrsmith35:
It looks like API 52.0 introduced a CloneSourceId reference in an SObject with a Type of null.  This fix avoids a CPE erroring out when trying to enter an Object>Field reference using fsc_flowCombobox.

PLEASE BUILD A NEW FlowScreenComponentsBasePack.


ericrsmith35:
This addresses #856 & #861

alexed1 (approved):

PR #874, opened 2021-08-21
Fixing property role bug

rygramer:
@jdayment @alexed1 when i was developing, i specified a role for these properties. because the original package didn't specify roles, there was an error when upgrading. clearly not the right thing to do.
reverting it back to the Josh's original specs. hopefully that does the trick.


alexed1:
Not a problem...the frustration here is that SFDC referential integrity won't let you make purely additive changes to those property elements. I don't see why that should be prevented.

rygramer:
Preventing the change of a role on a property seems like a good decision. If we change the specification from nothing to `inputOnly`, for example, users who are using the property as an output (which was allowed in a previous version of the component) are going to be SOL on upgrade.

alexed1:
Ooh good point

PR #888, opened 2021-09-01
Use setEntityAttachments() instead of setFileAttachments()

michael-eigenherd:
As a result of this change, the class was simplified quite a bit.

As far as I can tell in my tests, the attachment behaviour for users should be exactly the same.
The only difference seems to be, that attachments are now visible in activities.
Other than documentation and possibly apex tests (which all pass right now), I don't see any other impact of this change.

Unrelated, it seems to me, that `mapContentDocumentIdByContentVersion` will only ever have lists with up to `1` element.

### DIFF
--- a/flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmailAddAttachmentToEmail.cls
+++ b/flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmailAddAttachmentToEmail.cls
@@ -2,12 +2,13 @@
  * @description       : 
  * @author            : Refactored from Alex Edelstein
  * @group             : 
- * @last modified on  : 02-17-2021
- * @last modified by  : Jack D. Pond
+ * @last modified on  : 09-01-2021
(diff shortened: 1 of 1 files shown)


alexed1:
Nice! Welcome to the project, Michael. I love seeing new contributors. 
People are going to like that attachments are now visible in activities. @jdpond let me know if you have time to do a quick review.

jdpond:
@alexed1 I have merged @michael-eigenherd  pull request on https://github.com/jdpond/LightningFlowComponents/tree/master and tested it on that scratch org we've been using.  **It is an elegant and simple solution that solves a lot of issues**.  Wish we'd had it yesterday - so back to the cadence issue.

1. Should we add it to [sendBetterEmail 2.2.1](https://github.com/alexed1/LightningFlowComponents/milestone/4) or release immediately?
2. Would you like me to do a pull request from the now merged branched or just approve it directly from Michael's branch? (answer prolly depends on 1 above)

jdpond (approved):

Issue #900, opened 2021-09-09
SendBetterEmail Duplicate Accessor convertedFlowContext error


jimrae:
Just installed SendBetterEmail version 2.1.5 into a Winter 22 sandbox. When trying to add the action to my flow, I get the error "Error: Invalid observed convertedFlowContext field. Found a duplicate accessor with the same name."  In reviewing the javascript controller for the "sendBetterEmailCPE" LWC Component, I noticed that the variable "convertedFlowContext" is defined twice. Once on line 53 and a second time (with the @track annotation) on line 113. Removing the "untracked" version on line 53, allows the component to be configured correctly.

jdpond:
@jimrae Could you send me a screenshot of this error occurring?  Which browser are you using?

jimrae:
@jdpond here is the screenshot
![image](https://user-images.githubusercontent.com/1036171/132913343-10e067ad-ebb7-4175-a50a-f52404fd76f9.png)

And, here is where I commented out the duplicate variable to make the code work
![image](https://user-images.githubusercontent.com/1036171/132913411-823a06db-ae32-4780-8d95-a4ec158f7af4.png)

You will also see, the same variable existed again on line 113 with the @track annotation:
![image](https://user-images.githubusercontent.com/1036171/132913595-478e8354-78fb-43a2-9a7e-8591f6773c2c.png)

I am using Chrome on a Mac for my browser and installed directly from the Install packages link from UnofficalSF site.
As soon as I commented out the duplicate, the error went away, and the component worked correctly (and is awesome, by the way).

PR #905, opened 2021-09-10
 sendBetterEmail Duplicate Accessor convertedFlowContext Fix: #900

jdpond:
minor fix
sendBetterEmail Duplicate Accessor convertedFlowContext Fix: #900


alexed1:
Here's version 2.2.1 with these changes. https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G000003rUhSQAU

The general pattern I follow on version numbers is to increment the minor number if there are new features and the patch number if it's a bug fix release. But it's not set in stone

Can you smoke test the package and make sure it's looking as you expect?

Issue #902, opened 2021-09-10
CSV File Import - Two issues to report.


bottomlinebrokers:
I am using CSV File import v1.3.  I have two issues to report:

1 - I have comma characters in certain text fields that I am importing, therefore I need to change the delimiter.  But the only valid delimiter other than a comma character (from trial and error and viewing the code) is a semicolon character.  However, that means I cannot import and multi-select picklists, which naturally use semicolons to delimit multiple entries.  Am I doing something wrong, or do we need greater flexibility in what characters can be used as a delimiter?

2 - The documentation on https://unofficialsf.com/from-suraj-pilai-and-hijlko-csv-file-import-gets-some-great-internationalization-improvements states that if you leave the separators (field, thousand, and decimal) and currency symbol blank, it will default to existing default values.  However, (A) those fields are required in the Flow Action, and (B) when I leave one of those fields NULL, the code throws an error.

surajp:
@bottomlinebrokers The action should be handling commas within columns as long as you enclose the column in double quotes. you need not change the delimiter. You can see this [here](https://github.com/alexed1/LightningFlowComponents/blob/379a2f6efb4842e1b1c149548b3f08a3b464f360/flow_action_components/ConvertCSVToRecords/force-app/main/default/classes/ConvertCSVToRecordsTest.cls#L118) in the tests. Are these tests not passing for you?

bottomlinebrokers:
Suraj,

I tried that, but perhaps I didn't do it properly.


  1.  Please confirm that the columns need to look like:
  2.
  3.  ""Text, with comma"", ""next field"", ""field ; with ; semicolon"
  4.
  5.  Is it all or nothing?  In other words, do I only double quote the columns that need special handling, or do I have to double quote all columns within the CSV?

Once I get this response, I will try another test today and let you know.

Mark




________________________________

bottomlinebrokers:
Ok, apologies for the earlier comment.  I didn't read the test line carefully enough. :)  I ran a single record CSV upload test, and yes, it failed for my case.

Attached is a CSV with a single record to import.  The field in question that causes the crash is a Multi-Select Picklist named Real_Estate_Type__c, which has "Restrict picklist to the values defined in the value set" = TRUE for data consistency.  The picklist values are:

Agent, Commercial
Agent, Residential
Appraiser
Broker, Business
Escrow Officer
Landlord
Property Manager

[...]

PR #921, opened 2021-09-26
Fix issue with commas not being replace after parse

surajp:
Fix for issue #902 reported here.
The merged code was missing calls to add commas and quotes back in after
the values have been parsed.
The tests were missing asserts to validate that the final parsed value
matches what we expect when the value contains commas, newlines and
quotes. Added asserts to verify that.
I ran tests in #660 separately. However, did not commit those to the repo since those tests use Person Accounts.


alexed1 (approved):

alexed1:
sorry about the confusion. thanks for fixing the code up!

alexed1:
fixes incorporated into 1.3.2 https://unofficialsf.com/narender-singh-brings-csv-imports-to-flow/

PR #922, opened 2021-09-28
Added Default Values

ctalbotSFDC:
Adding ability to pass default values when navigating to an object if it is being opened in new mode
for NavigateEverywhereLFA aura LC


ctalbotSFDC:
missed one file

PR #923, opened 2021-09-28
NavigateEverywhereLFA - add default values

ctalbotSFDC:
Adds ability to include the default values via json as a parameter for navigating to an object's new record page

### DIFF
--- a/flow_action_components/MulticurrencyActions/force-app/main/default/lwc/jsconfig.json
+++ b/flow_action_components/MulticurrencyActions/force-app/main/default/lwc/jsconfig.json
@@ -1,0 +1,19 @@
+{
+    "compilerOptions": {
+        "experimentalDecorators": true
+    },
+    "include": [
+        "**/*",
+        "../../../../../../.sfdx/typings/lwc/**/*.d.ts"
+    ],
+    "paths": {
+        "c/*": [
+            "*"
+        ]
+    },
+    "typeAcquisition": {
+        "include": [
+            "jest"
+        ]
+    }
+}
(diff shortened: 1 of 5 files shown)


ctalbotSFDC:
Example for default value json (for defaulting contact fields) via a flow formula field:

'{"FirstName": "'  & {!firstname} & '", "LastName": " ' & {!Last_Name} & '", "Birthdate": "' & TEXT({!Birth_Date}) & '", "vlocity_ins__DriversLicenseNumber__c": "' &  {!Driver_s_License}   & '"}'

ericrsmith35:
This looks nice.  Do you have an update or blog post available to post on unofficialsf.com?

ctalbotSFDC:
> This looks nice. Do you have an update or blog post available to post on unofficialsf.com?

Thanks - it's just an update to the existing NavigateEverywhereLFA component that is already here for passing default values with the navigation.  It's like 3 lines of code added.

alexed1:
This is great! Christine, so is the use case that you can navigate to a new object page and prepopulate it with these defaults? I assume that if I don't provide a record Id but do provide an object name I get a new page?

alexed1:
It may just be 3 lines of code but it would be great to have a short post that shows it in action....

ctalbotSFDC:
Here's the info on it:

When using DestinationType = 'object', and DestinationAction='new', passing a json string of {FieldAPIName : value, FieldAPIName : value} will default those fields to the value.

Example:
To navigate to a new Contact page with defaulting the FirstName and the LastName fields via a formula field in the flow:

DestinationType = object
DestinationName = Contact
DestinationAction = new
defaultVals = {!formulafield}

{!formulafield} =
'{"FirstName": "' & {!firstname} & '", "LastName": " ' & {!Last_Name} &  '"}'

PR #928, opened 2021-10-02
ConvertCSVToRecords: Refactor and add more tests

surajp:
For ConvertCSVToRecords action:
Moved the statements for retrieving object fields map outside the inner
contentdocumentid loop for better performance. Added test to verify
handling of multiple content document ids. Added 'private' qualifier to
all methods in test class


alexed1 (approved):

alexed1:
fixes incorporated into 1.3.2 https://unofficialsf.com/narender-singh-brings-csv-imports-to-flow/

PR #930, opened 2021-10-04
Checked in 2 components

ctalbotSFDC:
getNextFiveRecords & getAggregateSOQL

### DIFF
--- a/flow_action_components/GetNextFiveRecords/force-app/main/default/classes/GetNextFiveRecords.cls
+++ b/flow_action_components/GetNextFiveRecords/force-app/main/default/classes/GetNextFiveRecords.cls
@@ -1,0 +1,93 @@
+public with sharing class GetNextFiveRecords {
+    
+    @InvocableMethod(label='Get Next 5 Records')
+    public static List<OutputVals> getRecords(List<InputVals> inputs) {
+        List<OutputVals> outputs = new List<OutputVals>();
+        OutputVals output = new OutputVals();
+        if (inputs.size() > 0) {
+            InputVals input = inputs[0];
+            if (input.listOfRecs != null && input.counter != null && input.listOfRecs.size() >= input.counter) {
+                output.totalNumRecs = input.listOfRecs.size();
+                integer currentCounter = input.counter;
+                integer endCounter = Math.min(currentCounter+5, input.listOfRecs.size()) - currentCounter;
+                output.newCounter = currentCounter + endCounter;
+                output.foundCount = endCounter;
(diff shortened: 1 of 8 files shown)


ctalbotSFDC:
GetNext5RecordsFlowAction
Apex Action that takes a position and a list, and returns the next 5 records from the list, starting at the position provided (or null if no records found) into separate variables

Add this apex action to a flow & pass the following input variables:

counter --> the index to retrieve the next 5 objects from the list
listOfRecs --> the list of objects to retrieve objects from
You will be able to retrieve the following outputs:

totalNumRecs --> how many items are in the list provided; -1 if list is empty
foundCount --> how many records were retrieved (0 to 5) since the list may not have 5 records to retrieve from the counter position; -1 if list is empty
newCounter --> next position in the list after retrieving the next 5; will be the last position of the list if foundCount is < 5; -1 if list is empty
percent --> gives the decimal percentage of the list you have perused; uses input counter / totalNumRecs; format is 0.10 for 10%
firstRec --> first object at position counter; is null if foundCount = 0 or -1
secondRec --> second object at position counter + 1; is null if foundCount < 1
thirdRec --> third object at position counter + 2; is null if foundCount < 2
fourthRec --> fourth object at position counter + 3; is null if foundCount < 3
fifthRec --> fifth object at position counter + 4; is null if foundCount < 4


AggregateSOQLFlowLightningComponent
Apex Action that allows a user to run an aggregate query and return a list of the results from one of the fields or expressions as a List of Strings

Add this apex action to a flow & pass the following input variables:

query --> the aggregate query to execute against the database
fieldname --> the fieldname from the query to retrieve and put into a list of strings
You will receive the following outputs:

List of Strings --> a list of strings that equates to the aggregate query's fieldname values
Example:

query = "select category__c, count(Id) from assessment_question__c where assessment__c = '0AB234001234567' group by category__c order by category__c" fieldname = "category__c"

results of query =

{"ABC", 4},

{"BCD", 2},

{"CDE", 3},

{"FGH", 10}

List of Strings = { "ABC", "BCD", "CDE", "FGH" }

alexed1:
Circling back on this....I think these are both highly useful extensions. I think that I'd like you to call the first one 'GetNextFiveRecordsIndividually' and add it to the collection processors project, and I think that the second one should be called MakeAggregateQuery and added to [this project](https://github.com/alexed1/LightningFlowComponents/tree/master/flow_action_components/ExecuteSOQLQuery/force-app/main/default/classes)
I know a lot of time has passed, Christine, since you offered this contribution, and you may have moved on to other things, but I would like to publish it if you're still interested.

Issue #937, opened 2021-10-08
ConvertCSVTo Records v1.3.2 - Fails to import dates in standard formats


bottomlinebrokers:
Importing fields on a record that are dates in the format MM-DD-YYYY and MM/DD/YYYY do not work.  There is no error message, the component does not fail.  The record imports successfully, the date fields are merely blank.

Please note that importing dates in the format YYYY-MM-DD IS successful.  I changed the dates in my test CSV (5 records only) to this format, and imported successfully.  However, that obviously is not practical on a large scale.  MM-DD-YYYY is the common U.S. format.

Mark

bottomlinebrokers:
Understood, but  be aware that MM-DD-YYYY is the primary format used throughout the United States, so there will be A LOT of Excel conversions.

That's okay for small data sets, but these conversions tend toward error prone for large data sets.  Is the date format that out of place.

Mark

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/ghei36>
________________________________

PR #938, opened 2021-10-10
Bug fix for different date formats

surajp:
There was a bug in the routine for handling multiple date formats. This is the fix for that. Also contains additional tests for testing different date formats in the csv.


surajp:
I originally had listed this as the fix for #937 , but this won't handle `MM-DD-YYYY` date format. That would require introduction of an additional date format input property, and I don't believe its worth complicating the action for that

surajp:
> Thanks, Suraj. I think it's worth adding a way for spreadsheets that have columns that are in MM-dd-yyyy will work here without modification. Is that something you could do?

Ok, @alexed1 . I will add an additional parameter for user to specify the date format (either MM-dd-yyyy or dd-MM-yyyy). If left blank, the format will be assumed to be either yyyy-MM-dd or the user's local date format in the org.  I will let you know when it's ready to merge

surajp:
@alexed1 any updates on when this might be merged?

alexed1 (approved):

alexed1:
I'll merge it now. Thanks for your patience and your fixes!

alexed1:
Posted as version 1.4...test it out if you get a chance. Thanks for your contributions again, Suraj!

PR #943, opened 2021-10-18
added reactivity by adding a setter for tableData that calls processD…

alexed1:
…atatable each time new data comes in


alexed1:
looks like there are some problems. closing

Issue #499, opened 2020-10-02
fbc_FormulaEvaluatorTest - Line 61 - Test evaluation should be '19' not '16'


SDJCorp:
https://github.com/alexed1/LightningFlowComponents/blob/9fc80b515558596d49b8ac93041ef04a7fb710e4/flow_process_components/FlowBaseComponents/force-app/main/default/classes/fbc_FormulaEvaluatorTest.cls#L61

This fails:
`        System.assertEquals('16', fbc_FormulaEvaluator.parseFormula('HOURS(DATETIME(2019,12,26,0,0,0))', stringContext));`

This Passes:
`        System.assertEquals('19', fbc_FormulaEvaluator.parseFormula('HOURS(DATETIME(2019,12,26,0,0,0))', stringContext));`

alexed1:
When I merged in this change the test started failing. Restoring the value to 16 made the test pass again. So, I'm confused. 

Reverting this for now.

SDJCorp:
Weird, maybe because of timezones or something? (the org I am using this on is based in Ireland)

alexed1:
That's a great hypothesis. Let's just comment out that assert for now until someone has time to dig into it.

adwhite75:
Closing this out as we have moved fbc over to a managed package which will alleviate any install/test class coverage issues.

PR #953, opened 2021-10-29
Fix tz issue in fbc_FormulaEvaluatorTest.cls #499

tomSny:
This PR corrects Unit Tests from failing for users not in the America/Los_Angeles

fbc_FormulaEvaluatorTest - Line 61 - Test evaluation should be '19' not '16' #499


alexed1:
Thanks for the fix, Tom. This is great.

PR #961, opened 2021-11-07
Added Hierarchy LWC component that shows parent and there children 

Sarveshgithub:
Below files are added in this PR
- hierarchy.lwc 
- HierarchyController.cls
- HierarchyControllerTest.cls

More Info added in Readme
https://github.com/Sarveshgithub/LightningFlowComponents/tree/master/flow_screen_components/hierarchy#readme


alexed1:
This is very cool, Sarvesh! Thanks for offering it to our repo!

Would it be possible for you to add to your documentation an example the use of this in a flow? Perhaps show the inputs and the result. A quick screencast is also great. The average flow user isn't as sophisticated technically as developers are so we like to to provide as much guidance as possible.

If you're interested, I'd like to invite you to expand your documentation into a blog post for publication at unofficialsf.com

alexed1:
@ericrsmith35 @adwhite75 fyi

adwhite75:
Ooo awesome! I've long wanted a flexible tree-grid component. Looking forward to trying it!

ericrsmith35:
Very nice.  Certainly a good addition.

Sarveshgithub:
@alexed1 I have updated the documentation with an example of Flow, Please check the latest [Document](https://github.com/Sarveshgithub/LightningFlowComponents/tree/master/flow_screen_components/hierarchy#readme).

alexed1:
Great additions! I have some more ideas here for making this as accessible as possible to non-developers:
1) what do you think of this idea: instead of requiring the user to create a perfect json structure, what if you just let them submit a comma-separated list of column names like this: 
"CaseNumber,Subject,Origin". Then, in your init code, you craft the json that lightning-tree-grid will expect from that list? That would remove the most complicated part of setting this up. 

2) that gif is very useful but it rendered too small to really see. Can you reshoot that with a tighter camera frame (i.e. zoom in)

3) I'm not clear on the relationship between parent field api name and the data that gets rendered. Is it that you start with the recordId and then find all the Object Name (Case) that have Parent Field equal to the recordId? Can you clarify that in the documentation? 

4) It appears that the initial version is read-only? Would it be possible to have the id of the record in any row that is selected output as an outputOnly attribute on the component? Then the flow could take subsequent action on the selection? 

5) can you add at the beginning of the documentation an example use case? For example , would this make a good one? "Suppose you want to navigate all the children cases of a particular parent case. Starting with the parent case, you can use this component to easily explore all of the children. Then you can select one of the children and click next in the flow screen and navigate to that record using Navigate Everywhere action"

alexed1:
Hi @Sarveshgithub, just checking in on this. It's ok if you don't have time to do those additional ideas, but if you're working on them, let me know so I don't publish your stuff prematurely.

Sarveshgithub:
Hi @alexed1 , Those ideas are on my todo list, I will try to complete them within this week. due to some busy schedule didn't get much time.

Sarveshgithub:
Hi @alexed1 ,

I have implemented the above ideas,
- Update columns with comma separated API name 
- Add checkbox to select the rows and store the selected ids in the collection
- Somehow I can't make a zoom in .gif format file, I tried many chrome extensions all are not good enough for zoom in .gif record :raised_eyebrow:

alexed1 (approved):

alexed1:
I believe that this top screen is no longer accurate now that you've added the nice ability to just pick fields. if so, can you remove: 
image

alexed1:
Regarding this statement: "Then you can select one of the children and click next in the flow screen and navigate to that record using Navigate Everywhere action", it looks like all of the attributes are marked role=inputOnly. So I wouldn't think a selection will be available. Did you get this to work? I definitely recommend you enhance the component to output a selected recordId.

alexed1:
published at https://unofficialsf.com/from-sarvesh-kumar-a-hierarchical-list-screen-component-for-flow/. Thanks again for the good addition...

PR #970, opened 2021-11-15
Datatable - Navigate to Next Flow Element on Save

iDamari:
Adding navigation on save to next flow element.
To be used with the Save Button to commit changes using a following Update Records flow element.

Idan.


ericrsmith35:
@iDamari I like what you have done here.  I will try and incorporate this in the next release.  Because there is an issue with how CPEs handle checkbox attributes, I'll need to update this to use the customized checkbox handling I've built for CPEs.

ericrsmith35:
This functionality will be included in the next version (v3.4.6)

PR #978, opened 2021-11-29
Convert csv to records

### LINKED ISSUE #971: ConvertCSVTo Records v1.4 - Fails to Import Negative numbers, Fails to Import Checkboxes as 1/0
Two issues with ConvertCSVToRecords v1.4 (most recent revision).

1 - I am attempting to import a negative value into a currency field "-2053".  The number imports as the positive number "2053". (Ignore the quotes).  I have not tried negative numbers into standard numbers fields.

2 - Attempting to import a column into a checkbox field with 1s and 0s representing TRUE and FALSE.  1/0 is how Salesforce exports checkbox from Reports.  However, when using 1/0, nothing is imported into the checkbox fields!  If I change the CSV data to "TRUE" or "FALSE", the field imports properly.  But when values are "1" or "0", no data is imported.

surajp:
Fixes #971 

- Combines the handlers for all numeric fields into one. 
- Handles negative numbers. 
- Handles '1' and '0' for 'true' and 'false' respectively for boolean fields.
- More tests for the above use cases


alexed1 (approved):

alexed1:
great fixes. new version 1.4.1 is up.

PR #974, opened 2021-11-23
https://developer.salesforce.com/forums/?id=906F00000008zoyIAA

hartnady:
https://developer.salesforce.com/forums/?id=906F00000008zoyIAA

Issue #942, opened 2021-10-18
*{!ComponentName}* - _({!ElementName})_ - {!BugTitle}


ericrsmith35:
Has anyone had this Test Class error when trying to install Execute SOQL into an org?

Test: ExecuteSOQLTest.testExecuteSOQLMonthDayCommaYear
Message: System.TypeException: Invalid date/time: 04/12/2020, 12:00 am

charles-thompson-salesforce-com:
Love the issue name, @ericrsmith35 !

PR #986, opened 2021-12-16
[FIX] ExecuteSOQL Datetime Parsing

sbresin:
## What does this implement/fix? Explain your changes.
There were 2 problems:
1. `Datetime.parse()` can only parse dates in the locale of the user, the unit test just assumed it will be en_US and failed for everyone else. (our default is de_DE for example, dates look completely different)

2. Salesforce is about to enable this critical update for everyone on January 31st: https://help.salesforce.com/s/articleView?id=release-notes.rn_forcecom_globalization_enable_icu_cruc.htm&type=5&release=224

Before this update `Datetime.parse()` could parse en_US dates without a comma (`04/12/2020 03:24 PM`) just fine, but not with a comma (`04/12/2020, 03:24 PM`)

**_After the update it is the other way around._**

The changes introduced with #974 made it work in orgs without the update enabled, but broke it for orgs with the update.
The scratchOrg that is used by the `force:package:version:create` command has it enabled by default, therefore the command fails.

In order to support both formats, i've added a regex to parse en_US dates regardless of the comma.
Also the monthDayCommaYear format is now parsed with a regex.

`Datetime.parse()` is kept as fallback for people using their locale formats.
**But If they have a multilingual org with users having different locales, it will still not work for them.** 

## Does this close any currently open issues?
#942 
issues introduced in #974 

## Any other comments?
IMHO it is not the best idea to support things that native SOQL doesn't. This way, the queries can't be copy-pasted in other tools and expected to just work.

Anyways some people in my team use the tool and had problems with failing tests.

**Also i think it should be added to the docs, that using locale formats is not recommended and users should stick to ISO dates.**

## Checklist
- [x] Code compiles correctly
- [x] Created tests which fail without the change (if possible)
- [x] All tests passing
- [ ] Extended the README / documentation, if necessary


alexed1:
This is a great writeup as it explains a topic very clearly that lots of us here in America don't understand well enough. Thanks for the fixes! Are there any other fixes or enhancements that should made to the SOQL Builder and Execute SOQL?

alexed1 (approved):

alexed1:
Regarding this:  "using locale formats is not recommended and users should stick to ISO dates."
as someone who does no work with locales, can you give an example of a user using a locale format? Is this something that would show up in the SOQL query?

PR #997, opened 2021-12-23
Fixes for UnlockRecord feature

isidle:
1)Switch to "without sharing" which allows non-admin users to use it (avoiding dreaded "Delete failed. First exception on row 0 with id x; first error: INSUFFICIENT_ACCESS_OR_READONLY, insufficient access rights on object id: []" error)
2)Bulkify input
3)Update API version to current stable (v53)


alexed1:
Hi, @isidle. Thanks for your contribution. Can you provide a little more detail on why With Sharing limits the use of this action to admins? My understanding from https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_classes_keywords_sharing.htm is that the class would use the context of the running user. What's the scenario where a non-admin can't use this?

isidle:
The "Modify all Data" permission is required in order to use the `System.Approval.unlock()` method.
https://help.salesforce.com/s/articleView?id=000321298&type=1

Thus, effectively the flow component will ONLY work if the running user is an administrator which significantly limits it's useful. 

This worked for any user in the past because of a bug in the flow engine, that executed everything as `system` ignoring the running users permission which was later corrected by one of the various flow critical updates. 

On the USF page, a number of people are complaining about this issue as they expect it to behave like it did previously, allowing non-admin users to unlock records (or allow flow triggers to do the same on their behalf). 
https://unofficialsf.com/approval-processes-unlock-records-in-lightning/

By changing it to  `without sharing` then any user will once again be capable of unlocking a record using this invokable action.

alexed1 (approved):

alexed1:
Great explanation. Thanks for improving this component.

alexed1:
I merged the code, but am running into this error when I try to package: (1) Apex Test Failure: Class.UnlockRecordTest.CanUnlockRecord: line 17, column 1 System.UnexpectedException: Apex approval lock/unlock api preference not enabled.

I've verified that the packaging org does have the preference enabled. I've put a question in to the approvals team.

isidle:
You might have figured this out but in you case you didn't have a chance to dig into it, it looks like one needs to Add "apexApprovalLockUnlock" under "orgPreferenceSettings"  in "project-scratch-def.json" to allow the package to be built successfully. 

{
  "orgName": "UnlockRecords",
  "edition": "Developer",
  "features": ["Workflow"],
  "settings": {
    "orgPreferenceSettings": {
      "s1DesktopEnabled": true,
      **"apexApprovalLockUnlock": true**
    }
  }

PR #999, opened 2021-12-28
Knowledge Recommendations Demo from JET BI

sahib-gadzhiev:
interim PR for Knowledge Recommendations Demo project


flow_action_components/ExecuteNBAStrategy/force-app/main/default/classes/ExecuteNBAStrategy.cls
```diff
@@ -130,4 +130,30 @@ public with sharing class ExecuteNBAStrategy {
...
+
+
+
+    @AuraEnabled
+    public static string launchFLow(Recommendation rec, String strategyName){
+        FlowDefinitionView flowDef = [SELECT Id, ProcessType, TriggerType, ApiName FROM FlowDefinitionView WHERE apiName = :rec.ActionReference];
+        if(flowDef.ProcessType == 'AutoLaunchedFlow') {
+            Flow.Interview curInterview = Flow.Interview.createInterview(rec.ActionReference, new Map<String, Object>{
+                'recordId' => rec.ExternalId,
+                'strategyName' => strategyName,
```
alexed1, line 149: this is only in here because of that recent issue where we need a dummy value, right? let's add a comment to that effect: //a limitation in the recordation interface means that a strategyName is still required to record a reaction even if the recommendation was generated by a flow and not a legacy-style strategy. Use any string that matches an existing strategy (the strategy won't be run and doesn't have to do anything)

flow_apps/Flow-Based Knowledge/force-app/main/default/classes/KnowledgeRecommendationsGenerator.cls
```diff
@@ -0,0 +1,91 @@
+public with sharing class KnowledgeRecommendationsGenerator 
+{
+    @InvocableMethod
+    public static List<List<Recommendation>> generateKnowledgeRecommendations(List<Requests> requests)
+    {
+        List<List<Recommendation>> recommendationsResult = new List<List<Recommendation>>();
+        List<Recommendation> recommendations = new List<Recommendation>();
+
+        String searchString = requests[0].searchString;
```
alexed1, line 9: the best practice for creating invocable actions is to bulkify them. that means that instead of just taking the first member, with [0], you iterate through the list. Most of the actions in the flowcomponents repo are properly bulkified like this. Here's one example: https://github.com/alexed1/LightningFlowComponents/blob/master/flow_action_components/AddQuotesToFields/force-app/main/default/classes/AddQuotesToFields.cls

When an action is bulkified, it can be used in flows that trigger on batches of records, like a bulk upload of Contacts, and the action will properly handle all of the records at once.

flow_apps/Flow-Based Knowledge/force-app/main/default/classes/KnowledgeRecommendationsGenerator.cls
```diff
@@ -0,0 +1,91 @@
...
+        Flow.Interview curInterview = Flow.Interview.createInterview(strategyName, new Map<String,Object>{'recordId' =>contextRecordId, 'searchString' => searchString });
+        curInterview.start();
+        System.debug('searchString' + searchString);
+        System.debug(curInterview.getVariableValue('outputRecommendations'));
+        return (List<Recommendation>)curInterview.getVariableValue('outputRecommendations');
+    }
+
+    public class InvocableErrorException extends Exception {}
+
+    public class Requests
```
alexed1, line 71: Just as you have a class here for Requests, you should use an inner class for Responses, even though this class currently only returns a single List of Recommendations. Doing this now makes it easy to later add additional outputs without affecting the signature of the class.

alexed1 (commented):
good start!

flow_apps/Flow-Based Knowledge/force-app/main/default/classes/KnowledgeRecommendationsGenerator.cls
sahib-gadzhiev: resolved it

flow_apps/Flow-Based Knowledge/force-app/main/default/classes/KnowledgeRecommendationsGenerator.cls
sahib-gadzhiev: resolved it

flow_action_components/ExecuteNBAStrategy/force-app/main/default/classes/ExecuteNBAStrategy.cls
sahib-gadzhiev: resolved it

alexed1 (approved):

PR #1025, opened 2022-01-20
Initial commit, InvocableStringWizard

munawirrahman:
Hi folks,

Adding new invocable method that will make all apex string methods available in flow. Also CPE already enabled in this method. would love your thoughts on this :)

thanks


ericrsmith35:
Do you have some documentation for this?

alexed1:
Mun, this looks promising! can you create a google doc that shows the inputs/outpus  and their function and allowable values and an explanation of the property editor controls?

alexed1:
or if you intend to publish it on your own blog, which is fine with me, can we see the draft so we can better understand the functionality?

munawirrahman:
hi @alexed1 @ericrsmith35 Draft of the Documentation shared to your emails ya.

> or if you intend to publish it on your own blog, which is fine with me, can we see the draft so we can better understand the functionality?

If you don't mind, I actually intend to publish this to unofficialsf

ericrsmith35:
Nice component.  I think a complete list of the available methods would be
a good update for the documentation.  A few more actual examples would be
helpful as well.

Eric

alexed1:
Here's a public copy of the draft post https://docs.google.com/document/d/1WQ0-BYFewHgnGqUbyAiY5lwDHg1Kn4XOuLepFZ7a79k/edit?usp=sharing

alexed1:
Mun, this is a REALLY great addition. I've wanted to expand the range of string processing for a long time. And I LOVE the CPE you've added. I'm happy to publish this action as it is now, but I have a few ideas for you to consider about how we might tune it.

1) Should this be aimed at developers or admins? 
right now, simply because you're exposing _all_ of the string methods directly, you're exposing the user to concepts like unicode and LevensteinDistance. Also, by exposing method names like charAt directly. That maximizes the power of the action but makes it intimidating to admins. By comparison, take a look at what I did at https://unofficialsf.com/easily-extract-substrings-tap-the-power-of-regex-with-findtext/ There, I'm using regex but I'm abstracting it into a simple 'Find String' concept for admins, so they don't even need to know that regex is doing the work. Of course, that means that that particular action can only tap into a fraction of the power of regex, but for me, that's the right tradeoff. My intent is that more regex functions will be exposed in this friendly way. 

There's also this 'String Converter' collection: https://unofficialsf.com/new-flow-actions-to-convert-csv-strings-to-string-collections-and-vice-versa/

Finally, a couple of years back I started a 'String Changers' package to do url encoding and value conversion: https://unofficialsf.com/flow-actions-for-strings-url-encoding-and-number-formatting/

In a situation like this, it's worth asking whether things can be consolidated so the user doesn't have to figure out which one to use and install multiple packages. 

As a thought experiment, I tried to imagine a single global 'StringWizard' solution that addresses all string modification issues. It would have your CPE but it would expose the modifications in an admin-friendly way and wouldn't send them to developer documentation (unless they indicated they wanted the 'advanced' UI info....).

I don't think that would require much modification from where you have StringWizard.... thoughts?

munawirrahman:
Hi Eric,

Thank you! your feedbacks are noted and i added the mapping to google spreadsheet in the documentation. Also for use case, I added a screen recording in the documentation as well.

Hi Alex,

I think thats a great thought, at first i was thinking to kind of embracing admin to understand more about the string concept as well (thats why i put the description and salesforce documentation url in the cpe), but i think i got your point to unifying every string processing methods into this action... technically that would be absolutely possible (we just need to add new private method and add another else if to the apex class, add new test, and add it to the CPE mapping). 

but, can we add that to the next release? i think we'll need to list which methods to be consolidated (is it only those 3 actions in your reply?) Also in the next release im thinking to add some kind of try/test mechanism in the CPE, the objective of this is to let admin/developer to explore/test their inputs and see what kind of result they can expect from this action (like what you did in HTTP Builder).

alexed1:
I tried installing and using String Wizard with the capitalize method, and got this unexplained error: 
image

alexed1:
1) For the documentation URL, I suggest you simplify it down to a single "Official Documentation" that's a URL, and not show the entire URL in the CPE. In general we try not to show URL's in consumer-facing UI.

2) I notice that the Method Description for charAt appears to actually be the description for center:
image

munawirrahman:
Hi Alex, 

can you retrieve the newest commit? it seems like you're still on the older one, there was one bug i fixed earlier

thanks

alexed1:
One thing that you might want to consider that could make a big difference in usability. Suppose that we generate a friendly phrase label for each of these methods. For example, the method **contains** would use the  label "Check target string for substring" and the method **join** would use the label "Convert target list into a single string".

alexed1:
Here are a couple of examples where I'd propose combining some similar methods into a single action:

Consider the methods based on contains:
image

Suppose that we aggregated this into a single 'Check for Substring' method. Instead of the user having to pick between contains and containsIgnoreCase, we add a checkbox IgnoreCase?  to Check for Substring. The code would use that checkbox to decide whether to call contains or containsIgnoreCase. To account for containsNone, we'd add a checkbox labelled 'Reverse Check (returns True if the substring is NOT present)'. etc....

munawirrahman:
@alexed1 static hyperlink text committed.

What do you think if i create another picklist for the custom methods that's not in the official documentation? the purpose is to avoid confusion to developers that doing things with flow and already familiar with the method label from sf official documentation?

1. Standard Method => for the methods that directly using method name from official docs, targeting developer
2. Custom Method => for the custom methods thats we re-label (or we provide to match more specific use case), targeting admin

-  https://unofficialsf.com/easily-extract-substrings-tap-the-power-of-regex-with-findtext/
- https://unofficialsf.com/new-flow-actions-to-convert-csv-strings-to-string-collections-and-vice-versa/
- https://unofficialsf.com/flow-actions-for-strings-url-encoding-and-number-formatting/
- Check for Substring
All those methods will be grouped into Custom Method (because we relabel it or its for specific use case)
if Standard Method selected, Custom Method will be disabled, vice versa. 
feel free if you have suggestion for the picklist label

alexed1:
I think it has to either be a developer-targeted action or an admin-targeted action. I think the relabelling and aggregation of multiple methods into a single picklist choice only makes sense if it's an admin-targeted action. It's ok to keep it developer-focused. Happy to publish it as is. But my preference would be an admin targeted UI that had only friendly labels

alexed1 (approved):

alexed1:
Here's a version 1.0 of Invocable String Wizard. @munawirrahman can you check it out and verify that it's behaving properly https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G000003rUqLQAU   ?

munawirrahman:
Hi Alex,

I have checked it, and I think its already behave properly

thanks!

PR #1031, opened 2022-01-28
General Improvements to Quick Record from JET BI

sahib-gadzhiev:


alexed1 (approved):

flow_screen_components/datatable/force-app/main/default/lwc/datatable/datatable.js
```diff
@@ -263,6 +263,7 @@ export default class Datatable extends LightningElement {
     @track isAllFilter = false;
     @track showClearButton = false;
     @track tableHeightAttribute = 'height:';
+    @track tableBorderStyle = 'border-left: var(--lwc-borderWidthThin,1px) solid var(--lwc-colorBorder,rgb(229, 229, 229))';
```
alexed1, line 266: @ericrsmith35 this was added in response to a request of mine to create a visible left border....

ericrsmith35: Thanks Alex.  Can you think of any reason this should be optional or selectable?

alexed1: not so far...i guess we keep our eyes out for a scenario where it doesn't look good. If so, we make it optional

I also just asked him to do the same thing for the tob border and right border 
image

ericrsmith35:
The outer border currently around the datatable is optional and with your
updates, the table may look better without that extra border.

PR #1041, opened 2022-02-04
Encode flow params using encodeURIComponent instead encodeURI

JeroenSfdc:
If params would include e.g. value containing URL character, it might not be sufficient to use encoreURI.
Assume you have to encode a param like `inputDateTime=2022-10-02T08:30:00+01:00` only encoreURIComponent will encode the '+'.


alexed1 (approved):
Thanks for the fix, Jeroen!

PR #1047, opened 2022-02-11
Add Safe Navigation Operator to converted sObject

RealEvanPonter:
In a Multi-Currency Organization, if you are using Datatable to define new records (by placing a temporary unique value in the Id field that starts with the 3-character object prefix), a null pointer exception is thrown at line 321 when attempting to access the value of the currency field from the sObject retrieved via SOQL query (which found no results). I added a Safe Navigation Operator at line 321 to ensure the "converted" variable is not null before attempting to retrieve the value of one of its properties.


ericrsmith35:
Thanks, I will incorporate this in the next release.

ericrsmith35:
This will be part of v3.4.5.  I will later be adding an option to bypass the currency conversions completely.

PR #1062, opened 2022-02-25
Quick Query Notes 2-18 from JET BI

sahib-gadzhiev:


flow_screen_components/datatable/force-app/main/default/lwc/datatable/datatable.js
```diff
@@ -267,8 +267,8 @@ export default class Datatable extends LightningElement {
     @api types = [];
         
     // Other Datatable attributes
-    @api sortedBy;
-    @api sortDirection; 
+    @api sortedBy = '';
```
alexed1, line 270: @ericrsmith35 note these proposed changes to Datatable. They were added in response to 

_5.3 manual changes to sorting need to be reflected in the configuration

I clicked on the column header for Subject, causing the rows to change their sort order to sort by Subject, Descending. I then re-entered the Config menu. 
Expected: I should see a single sort key set with Subject, Descending_

Let me know if you have any concerns with this approach. An alternative would be to have Datatable output a [TableViewDefinition record ]. That would provide a more general solution for capturing any changes made manually at run time. however, most of the view definition can't be altered at runtime. If it's worth capturing the sort order, it's probably worth capturing changes to the width of the columns. (I'm assuming that such changes are detectable by the component code). but maybe we just add that as a discrete output, like Sahib has done here with sortedBy and sortedDirection, instead of add the complexity of the view definition, which is mostly design time relevant. Thoughts?

alexed1 (approved):

flow_screen_components/datatable/force-app/main/default/lwc/datatable/datatable.js
ericrsmith35: I wouldn't suggest option 2 (changing the definition).  I feel that user interaction with the table @ runtime is one of the benefits of the component but that shouldn't change the original configuration the designer provided when creating the flow.

alexed1: When I try to install this code, @sahib-gadzhiev , I'm getting this error:
 Encountered errors installing the package!,Installation errors: 
1) (datatable) lwc/datatable/datatable.js-meta.xml: You can’t remove the following public properties: sortDirection, because the component is part of a managed package., Details: datatable: lwc/datatable/datatable.js-meta.xml: You can’t remove the following public properties: sortDirection, because the component is part of a managed package.

it looks like sortDirection has been in the component a long time. I'm not sure why 'managed package' is relevant here though. 

It looks like you replaced sortDirection with sortedDirection, so the functionality didn't really change. Can you change sortedDirection back to sortDirection?

PR #1066, opened 2022-03-02
Handle null value

askelia:
When multipicklist field has no item selected, its value is null. 
So the attempt to execute `fieldValue.split(';')` will raise an exception. If the value is null then returning an empty list makes sense.


alexed1 (approved):
Thank you very much for this good enhancement. Glad someone is thinking about the nulls =)

PR #1068, opened 2022-03-03
Adding textAreaPlus

jdayment:
Worked with Nathan Shulman from SF on this one for another user looking to do this in flow from a post in OhanaSlack basically a way to display character count and validation on a plain text input field. I'll work with Nathan on a post. We have some ideas for more things to do on it.


alexed1 (approved):
looks like a great addition!

alexed1:
I assume you'd like a package of it now?

jdayment:
Yes I will get a post written up this weekend.

alexed1:
Here's 1.0.0.0 https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G000003rUsWQAU

PR #1076, opened 2022-03-09
jerrypoon - Fixed content overflowing from table when table height is set

JerryPoon:
- Removed extra div. This was overflowing as the height is dictated by the parent div
- Moved border styling to CSS
- Removed padding by removing slds-box. This is replaced by custom styling of border from previous commits


JerryPoon:
Note. The current proposed changes look like this (no padding)
image

If we were to keep slds-box, then the height attribute needs to be set on the inner div. it'd look like this
image

ericrsmith35:
I will review this and pull these changes into my v3.4.6 release.

ericrsmith35:
To be included in v3.5.0

PR #1086, opened 2022-03-23
NetworkId add

cadillacben:
I'm only a beginner developer and have almost no idea how to use github, so we can start with that.  :)

I was trying to use this for community users and found I needed to add the NetworkId to make this work.  All I did was to add that as an input and add the bit below.  In my flow I get the $User's NetworkMember.NetworkId to use as the input, but there could be a better way to do that.

```
if(networkId != null ){
                             cv.NetworkId = networkId;
                            }
```

The test class does need to be updated to pass in the networkId string, but can just be '' from what I found.

### DIFF
--- a/flow_action_components/FlowActionsBasePack/force-app/main/default/classes/ConvertRecordsToCSVQueueable.cls
+++ b/flow_action_components/FlowActionsBasePack/force-app/main/default/classes/ConvertRecordsToCSVQueueable.cls
@@ -1,3 +1,16 @@
+/**
+ * Copyright (c) 2020 CapTech Consulting
+ * @File Name          : ConvertRecordsToCSVQueueable.cls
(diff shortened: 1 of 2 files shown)


alexed1:
Hey! Thanks so much for this contribution!

I'm puzzled why this action, which simply does some text parsing, would require a NetworkId... I'd like to look at your specific code changes, but if you look at https://github.com/alexed1/LightningFlowComponents/pull/1086/files you can see that it's showing enormous amounts of change. So something is a little wrong with your pull request. 

I encourage you to close that pull request and try creating a new one. What you're looking for is where only the specific changes you made show up in color. 

You might want to look online for some tutorials on creating pull requests. It's tricky stuff when you haven't done it before, but rewarding if you bull through it.

PR #1099, opened 2022-04-12
ConvertCSVToRecords - fixes #1100 - refactor response aggregation to avoid bulk execution list size mismatch error from Flow engine

### LINKED ISSUE #1100: *ConvertCSVToRecords* - _(ConvertCSVToRecords.cls)_ - multiple file handling results in request/response size mismatch error in Flow engine
### Steps to reproduce

1. Add element to screen flow.
2. Provide multiple input CSV files
3. Attempt to convert
4. Flow engine returns error

### Expected behaviour
Invokable action should create a single response object to match the single request object call to this invokable method. It should parse several input files into one aggregated list of SObjects and populate the `convertedCSVRows `variable in a single Response object 

### Actual behaviour
Method interprets the multiple files as a desire to create multiple responses, rather than parse and aggregate all the rows found in all the input files into one list in a single `Response`, a separate `Response `object is created for each file. This breaks the specification of the Flow engine for an invokable to return the same number of `Response `object as were provides as `Request `objects.

### Debug Logs
Error Occurred: The number of results does not match the number of interviews that were executed in a single bulk execution request.

kurtfairfield:
Getting Input and Output collection size mismatch errors with multiple files. Updated to pass back only one Response in the List, as that is what is passed in. Updated unit tests as well to match

### DIFF
--- a/flow_action_components/ConvertCSVToRecords/force-app/main/default/classes/ConvertCSVToRecords.cls
+++ b/flow_action_components/ConvertCSVToRecords/force-app/main/default/classes/ConvertCSVToRecords.cls
@@ -9,6 +9,8 @@ public with sharing class ConvertCSVToRecords {
     Response[] responseList = new List<Response>{};
 
     for (Request curInput : inputs) {
+      Response output = new Response();
+      output.convertedCSVRows = new List<SObject>{};
       if (curInput.contentDocumentId == null && curInput.contentDocumentIdList == null) {
         throw new FlowCustomException(
           'You need to provide either a single ContentDocument Id OR a list of ContentDocumentIds as inputs to this Flow action.'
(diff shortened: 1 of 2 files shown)


kurtfairfield:
Error Occurred: The number of results does not match the number of interviews that were executed in a single bulk execution request.

Code was creating a unique Response object for each file, rather we are aggregating multipl files into one list and passing those coverted rows back in a single Response. List of Responses is to handle bulkificaiton of calls to this method, not parse each uploaded file into its own response. This causes input/output size mismatch error in flow engine.

kurtfairfield:
Still working on this particular invokable action to make it handle larger files. Even after this fix was implemented so we can use multiple files, if the file is too large, the parser hits heap size governor limits, likely in the recursive `removeFSeps()` method call.

Work around is to break the large input file (in this case a Custom object with about 7x 80 char fields typically) we can only parse 2500 rows before heap limit hit. But with several separate paged files, the parser survives. 

Will likely come back to this and try and get it to handle much larger input files.

alexed1:
Thanks for the good fix and welcome to the project! Yeah, if you end up having time, it would be lovely to get some smart handling for larger inputs. That's an area most of us don't have much dev experience at.

alexed1 (approved):

PR #1115, opened 2022-05-18
Screen Flow Components - Refresh View and Navigate to Next

mbiondis:
Users may use a Quick Action on a record page that calls a **screen flow that performs tasks that don't require user intervention** (e.g. edit fields or create related records). Without these components, the user has to close the modal (by clicking next, or with the X) and refresh the browser to see the update.

- The "Navigate to Next" screen component automatically "finishes" the flow and closes the modal.
- The "Refresh View" updates the current view without reloading the browser window.

You can get details of the use case and reference material in the readme file of each component. 
Below is an example of how to combine the two with the Show Toast in a screen flow.

FlowActions_Example

### DIFF
--- a/flow_action_components/RefreshView/.forceignore
+++ b/flow_action_components/RefreshView/.forceignore
@@ -1,0 +1,3 @@
+# List files or directories below to ignore them when running force:source:push, force:source:pull, and force:source:status
+# More information: https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_exclude_source.htm
+#
--- a/flow_action_components/RefreshView/README.md
+++ b/flow_action_components/RefreshView/README.md
@@ -1,0 +1,21 @@
+# Refresh View
+
+## Use Case
+
+When a flow makes changes to record fields, you want your users to see those changes without them having to reload the browser page manually.
+
+## How to install
+
+Deploy to an org using:
+
+```
+sfdx force:source:deploy -p flow_action_components/RefreshView/force-app/main/default/aura/refreshView
+```
+
+## How to use
+
+Simply add this invocable action at the end of your flow to refresh the view. This is especially useful in screen flows after your last screen.
+
+## How it works
+
+This invocable action uses the `force:refreshView` event to reload the current view automatically. Refer to the [Salesforce Developer Documentation](https://developer.salesforce.com/docs/component-library/bundle/force:refreshView/documentation) for the techincal details
--- a/flow_action_components/RefreshView/config/project-scratch-def.json
+++ b/flow_action_components/RefreshView/config/project-scratch-def.json
@@ -1,0 +1,9 @@
+{
+  "orgName": "Demo Company",
+  "edition": "Developer",
+  "settings": {
+    "lightningExperienceSettings": {
+      "enableS1DesktopEnabled": true
+    }
+  }
+}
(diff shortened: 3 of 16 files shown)


alexed1:
Thanks so much for this contribution, Martin. You've identified a gap and come up with a solution. Really appreciate this. 

There will be questions about how your Navigate to Next is different from https://unofficialsf.com/navigate-everywhere-flow-action/...what's your take on that? Could we combine the two into a single navigate solution? 

Likewise

alexed1:
Hi, Martin. Any thoughts on the above?

mbiondis:
https://blog.texei.com/refresh-a-record-page-from-a-lightning-web-component-14a5874ff68e

alexed1:
@ericrsmith35 @adwhite75 interested in your thoughts on how to position this useful new functionality from Martin.

let me try one out:
Navigate Everywhere is a navigation action that you can add to screen flows to direct the screen flow to a wide range of destinations. Your browser will load the appropriate URL when the flow reaches this action.

NavigateToNext is a screen component that triggers a Next navigation as soon as it loads. If you put this component on a screen and put that screen at the end of your screen flow it will have the effect of closing the flow, instead of leaving the user with the Your Flow Finished screen.

Update Screen is a local action that calls the Salesforce recordData component when it loads, which causes a refresh of that record. 

Refresh View is a local action  that calls the Salesforce refreshView component. This also causes a refresh. Martin asserts that Update Screen doesn't refresh all of the components. Should we retire Update Screen in favor of Refresh View? I'm happy to do so if it makes sense...

ericrsmith35:
I like what he has done here.  I agree with replacing Update Screen with
Refresh View.

I am going to test the NavigateToNext to see if it will close my Column
Wizard flow that runs in a modal in my Datatable CPE.

Eric

ericrsmith35:
It doesn't look like these components follow standard aura component naming conventions.  Is that an issue?

ericrsmith35:
I can't test the navigateToNext component because it is not exposed to flow.

ericrsmith35:
I was able to get the component installed by renaming the .js and .xml code pieces.  The navigateToNext failes when added to a screen that is the last element in the flow.
image

ericrsmith35:
Changing the action to FINISH and testing on my Datatable Configuration Wizard flow, which runs in a modal, the component just causes my flow to restart.  It does not close the modal.

alexed1:
@mbiondis , what are your thoughts on @ericrsmith35 's findings?

alexed1:
Hi, @mbiondis, checking in on this....

mbiondis:
Hi @alexed1, really sorry for the delay, I was away on annual leave for the last month-and-a-half.

@ericrsmith35 I think the behaviour makes sense, because the flow you're testing on requires the user to exit the modal with the X button, rather than with a 'Finish' button. The intended purpose of this component was to auto-close screen flows triggered from quick actions and URLs, but not from a CPE.

@alexed1's summary from July 1st is spot on as to why I created these components, as I felt the existing alternatives weren't meeting the requirements I needed.

alexed1 (approved):

alexed1:
As Eric pointed out, this code would need to make a distinction between NEXT and FINISH. If you put this component on the last screen of a flow it will throw an error because NEXT is not an allowable option.

alexed1:
So that I can explain it, I want to better understand the difference between the force:recordData (which is what UpdateScreen uses) and florce:refreshView, which is provided here.

I notice that the documentation for refreshView says 'it's preferable to use recordData' down at the bottom of the page (https://developer.salesforce.com/docs/component-library/bundle/force:refreshView/documentation)

I'm happy to have both in the repo if they're not overlapping. What's the advantage as you see it of refreshView?

PR #1116, opened 2022-05-22
Added spinner on save to Record Detail FSC

dannysummerlin:
Just added a custom onsubmit handler to show a spinner as soon as the submission has started so users have a better sense of what is happening.


alexed1:
Thanks for the nice refinement, Danny! Welcome to the FlowComponents repo.

I'm getting a merge conflict: 
image

Should I assume that the correct version is:
this.recordId = event.detail.id;
?

dannysummerlin:
Hi Alex! I went ahead and did the resolution - the this.isSaving is the correct version, it needs that to clear the spinner when the save is done! Hope that clears things up, lemme know if it needs anything else

alexed1 (approved):

PR #1139, opened 2022-06-15
Adds useful things to  the GetPicklistValuesBulkified

renatoliveira:
* Adds prettier to the project in this folder.
* Adds logging level to debug statements.
* Adds a 'ok' and 'message' field to the response so that the failure can be handled gracefully within the flow.
* Adds labels to the return value.
* Adds extra test and renames them to both situations they test.

I've seen the message on SFXD to add the record type option (which explains my branch's name), but that is not possible without making a callout, as pointed by other users (for now, that is: https://ideas.salesforce.com/s/idea/a0B8W00000GdVwoUAF/getting-picklist-values-based-on-record-type).

Meanwhile I've added some things to the action that might be useful to the end user.


flow_action_components/GetPicklistValues/yarn.lock
```diff
@@ -0,0 +1,221 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
alexed1, line 1: Is this file needed? In the past, we've had maintenance headaches from lock files of this site when the versions change, security issues are found, etc...

alexed1: Thanks so much for adding these improvements, @renatoliveira!

alexed1: my preference would be to remove the prettier aspect. We've just had too many little problems with references to external libraries and lock files and I'd like to keep the repo internally consistent where possible.

renatoliveira: Removed the file. Maybe it is a good idea to add a "contributing" section where this is noted so other people don't try and commit it the same way I did initially.

alexed1: You're absolutely right. I wish I had a couple of volunteers to tune our content for developers interested in contributing...

I'll get this merged and packaged soon

PR #1146, opened 2022-06-22
Version upgraded to handle NetZero Cloud objects

baslu93:
The quickLookup component was throwing the following error if used on "Stationary Asset Environmental Source" (NetZero Cloud object)

image

with the newer Apex class API version everything works fine.


ericrsmith35:
@alexed1 I've created the v1.0.4 package and updated the documentation page.  Please process PR#1153

PR #1151, opened 2022-06-28
Update fsc_screenFlow.js

jdayment:
Fixes issue for orgs with enhanced domain enabled where url for flow was returning as null due to looking for force.com in url so it wasn't working in vf page, experience site, or if on an external site.


alexed1:
Nice fix, Josh!

alexed1 (approved):

PR #1164, opened 2022-07-08
Changes in Case the we just want to pass the recordID

Adityaonmove:
Added 2 more variables

        @InvocableVariable
        public String recordId;
        
        @InvocableVariable
        public String objName;

and a method to get the Approver Id


alexed1:
Nice improvement, Aditya!

PR #1182, opened 2022-08-04
Trim spaces from split csv string

Tamarer:


alexed1:
Let me know if you'd like to reopen and discuss this one, @Tamarer.

Tamarer:
@alexed1  can you please explain why this update can’t be merged. I may have misunderstood If I had some additional tasks here, thanks!

alexed1:
I don't think the code you're proposing here will work because I don't think you can directly access csvString.

Tamarer:
I think I can access csvString because it’s declared in line 9 but for some reason this update doesn’t actually trim the spaces.
Thank you for your help, you can delete this branch.

PR #1186, opened 2022-08-15
Add flowAutoNavigate

jdayment:
Flow Auto Navigate does exactly that when added to a screen flow the admin/developer can have the flow auto navigate to the next action or screen after a set period of time. The timer can be displayed or not displayed and when displayed there is an option to reset the timer for the end user if you decide to display it. 

I will work on getting a post put together.


alexed1:
Looks great. I'll package as soon as you create the post.

alexed1:
v 1.0 https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G0000043wpcQAA
I can't remember, do you have access to post yourself? I can do it....

jdayment:
I do I'll get it updated

PR #1188, opened 2022-08-23
sendBetterEmail - add MessageId support

jeff-restlesslabs:
modified:   flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmail.cls
modified:   flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmailRequest.cls
modified:   flow_action_components/SendBetterEmail/force-app/main/default/flows/Send_Better_Email_Testflow.flow-meta.xml
modified:   flow_action_components/SendBetterEmail/force-app/main/default/lwc/sendBetterEmailCPE/sendBetterEmailCPE.html
modified:   flow_action_components/SendBetterEmail/force-app/main/default/lwc/sendBetterEmailCPE/sendBetterEmailCPE.js

### DIFF
--- a/flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmail.cls
+++ b/flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmail.cls
@@ -131,6 +131,8 @@ public without sharing class SendBetterEmail {
       String replyEmailAddress = curRequest.replyEmailAddress;
       String senderDisplayName = curRequest.senderDisplayName;
       String subject = curRequest.subject;
+      String messageID = curRequest.messageID;
+        System.debug('Message ID = ' + messageID);
       String templateID = curRequest.templateID;
       String templateName = curRequest.templateName;
       String templateLanguage = curRequest.templateLanguage;
@@ -345,6 +347,9 @@ public without sharing class SendBetterEmail {
 
         // Specify the subject line for your email address.
         mail.setSubject(subject);
+
+         // Specify the Message ID.
+         mail.setReferences(messageID);
 
         // Set to True if you want to BCC yourself on the email.
         mail.setBccSender(bcc);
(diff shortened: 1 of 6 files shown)


jeff-restlesslabs:
I have updated the necessary components to support sending Message Id's for email to Case responses.  
When you create an email via an SF Email Alert the system generates Message Id which is stamped in the Reference of the outgoing email.
When there is a reply to this Alert generated email the Email-To-Case service will match the incoming email to the Case based on this Message Id. Unfortunately i cannot find anywhere online regarding the format of these Message Id's required for this default matching.
However there is a workaround...
If the Message Id on incoming email matches an Email Message record (02s Object) which is related to the Case then Email-To-Case will match it to a Case - the format does not matter the Message Id on the incoming just needs to match that of an Email Message.

flow_action_components/SendBetterEmail/force-app/main/default/classes/SendBetterEmail.cls
```diff
@@ -131,6 +131,8 @@ public without sharing class SendBetterEmail {
       String replyEmailAddress = curRequest.replyEmailAddress;
       String senderDisplayName = curRequest.senderDisplayName;
       String subject = curRequest.subject;
+      String messageID = curRequest.messageID;
+        System.debug('Message ID = ' + messageID);
```
jdpond, line 135: Prolly want to delete this line? [SendBetterEmail.cls]135 (don't need the debug statement in prod?)

jeff-restlesslabs: Good point

jdpond (commented):
Hey, I looked a this and it seems to be OK, but I didn't do a full test.  Can we add @jcallahan526 to the alexd1 LightningFlowComponents so we can document/share this as an issue?

Also, @jcallahan526 I didn't load this a run a test of all the test flows in Send_Better_Email_Testflow Flow - won't be able to until this weekend, but it all looks good - just haven't tested.

Finally, and this is because I just saw this today: https://trailhead.salesforce.com/trailblazer-community/feed/0D54S00000BshKK

I'm guessing what is happening is when the BC and CC fields are deleted in the editor, they are being reverted back to "" instead of _null_  .  Could fix it either by modifying the editor or (perhaps more simply) modifying the class to see "" as null.

Wanna take a look at that too?

jeff-restlesslabs (commented):
@jdpond I removed the debug line from line 135 of SendBetterEmail

alexed1 (approved):
Great addition, Jeff. Thanks for helping out with and improving this action.

PR #1189, opened 2022-08-25
Example of Flow from Find Object Type from Any Id Post

jkranz-rk:
This is an example of the Flow that is discussed in the post on UnofficialSF, "Find Object Type from any Id in Flows"


alexed1 (approved):

alexed1:
Alexed1 to package

PR #1193, opened 2022-09-01
Rename TemplateForInvocableActions to TemplateForInvocableActions.cls

claudiojp94:


alexed1:
Thanks for the contribution, Claudio! I'm actually going to remove that file now that you've pointed it out because it has some antipatterns in it (the use of [0] instead of bulkifying, namely)

alexed1 (approved):

claudiojp94:
> Thanks for the contribution, Claudio! I'm actually going to remove that file now that you've pointed it out because it has some antipatterns in it (the use of [0] instead of bulkifying, namely)

Yeah, you should also update the unofficialsf package of UnlockRecords, It's still on an old version, that hasn't got the change to public without sharing to allow non-admin users to perform the action.

PR #1201, opened 2022-09-07
Detect and Launch - Launch on Field Change

AndyHaas:
These changes will allow the user to enter a field to watch and a field value that the field to watch must equal. If the field to watch matches the field value, then we will launch the Edit Flow. 

If both field to watch and field value are not specified, then the component will act as normal.

I will write up the change doc and how to customize the new component in the following comment.


AndyHaas:
@alexed1 and @ericrsmith35 
Here is the documentation. Feel free to clean it up, and if you have any questions, please let me know. 

https://docs.google.com/document/d/16MOYyQS56GQXJuMSAYlAYkMqZYXM0tgkwDA-PYs56o8/edit?usp=sharing

alexed1:
This looks like a very meaningful enhancement, Andy. I've never given force:recordData the attention it deserves, but it looks like you've figured out how to use it to essentially trigger screen flows on record changes, (which is probably how I'll title the blog post).

A couple of requests:
1) if you can, record a simple screencast showing it in action.
2) let's reorg your documentation a bit. In your first paragraph, you end up getting into some fine details, while your most important text is the sample use case that's near the bottom. Let's get that use case up at the top.
3) If "ID" is a must have, perhaps prepend it automatically so the user doesn't have to add it themselves?

alexed1:
@adwhite75 check it out! Triggered screen flows!

AndyHaas:
@alexed1 
Thank you for the feedback about the Id field. I have made the necessary changes to have the ID field auto-populate.

I will also get those screen recordings and send them over to you.

alexed1 (approved):

PR #1203, opened 2022-09-07
Winter 23 Fix - fire attributes

ericrsmith35:
The fire attributes stopped being sent back to the flow in Winter 23.  This fixes that issue.

@alexed1 Please repackage and send me a confirmation so I can update the unofficialsf pages.


alexed1:
will do. Once this bug is fixed for Winter 23, will we have to make another change here?

alexed1:
on my initial search, it looks like this was never packaged. I also can't locate a blog post about it. VS Code says you touched it in 2018 and I touched it '4 years ago'. Did you spot a post where we talk about it?

ericrsmith35:
I have created a new package and updated the documentation.  I'll send another PR with a minor change to the test class

ericrsmith35:
This will not require any changes after a W23 "bug fix" is completed.

PR #1209, opened 2022-09-09
Update FindCommonAndUniqueRecords.cls to optionally return null

KNLangston:
add extra input option to FindCommonAndUniqueRecords that allows it to return empty Lists as null instead


alexed1:
Thanks for the PR, Keenan. Can you tell me a little more about this? Should we always return empty lists as null (i.e. was this a bug?) or is there a case for both to exist?

KNLangston:
Hi Alex,

I made this change to allow flow Decision elements to use the "Is Null" operator on the outputs of this invocable action and have the outcome be True when no records are returned.  When the action returns empty Lists, "Is Null" returns false even when the List size is zero.  I find this easier and I thought other users may also appreciate this convenience, so I submitted a pull request.

The reason I made it optional was so that previous implementations of the action that are expecting a non-null return value will not experience errors.

Thanks,
Keenan

alexed1 (approved):

PR #1218, opened 2022-09-15
Added Test Class for Apex Class: Resolve Approval Requests

AndyHaas:
This test class will help pass the Resolve Approval Requests to production. Unfortunately, a lot of hard coding had to be done as Process Instances, and Process Definitions can not have a DML insert function. 

When using this test class, you will need to create the following:

- An Account with the name "Acme (Sample)."
- An Approval Process on the Account object with a Developer Name "Test_Process_Definition."
- Activate the Approval Process
- Submit for Approval the Acme (Sample) account. Leave it as pending, as that is what the code looks for.

If you want, you could make these changes within the test class to fit your orgs' test records, but this will need to be done for this class to run. You then will get a 93% code coverage.


alexed1 (approved):
Thanks, Andy!. This makes this utility much more deployable.

Issue #1211, opened 2022-09-11
Customisable Banner - please include font weight


JodieM:
**Describe the solution you'd like**
I'd really like font weight to be an option on the banner

**Additional context**
I am trying to get the flow to look as close as possible to standard related record components, and without font weight I can't make the header bold enough to look like standard

alexed1:
Hi, Jodie...can you clarify which component you're referring to? [Is it this one?](https://unofficialsf.com/from-sharad-style-your-flow-with-headers-banners-rules-formatting-and-more/)

JodieM:
Yes that's the one, thanks @alexed1

alexed1:
Fixed by Andy Haas in version 1.0.1 https://unofficialsf.com/from-sharad-style-your-flow-with-headers-banners-rules-formatting-and-more/

PR #1220, opened 2022-09-16
Updated Custom Header to support Font Weight

AndyHaas:
If the Bold checkbox is checked, you can choose a font weight. If a font-weight is not set, the font weight will be set to bold. Anyone using this component and updates won't break as this is current functionality. 

Once you set a weight, you won't be able to undo a weight. A normal bold is 700. 

This is for #1211


alexed1 (approved):

alexed1:
@AndyHaas regarding 'you won't be able to undo a weight', are you saying that the user can't change the value after they pick it the first time?

Issue #962, opened 2021-11-08
Quick Lookup - Child filter not respecting Parent Default Value 


CultechPtyLtd:
On your instructions you give 2 options for using default values. For Option Two, it's a matter of also populating the O1_Output Value. This works fine when you are not using any Child Lookups off the Parent. However, If you are using a child lookup then that lookup does not respect the populated defaulted value of the parent it is using for filtering. You have to clear the parent and reselect it in order for the Child to begin its filtering on that value.

Because of this, I am forced to use the Option One for the default value which is not ideal as its more steps for the user to complete.

Would be great to have the Child respected the Pre-populated/Default Output of the Parent.

CultechPtyLtd:
This sort of ties into this bug I posted a while back:

https://github.com/alexed1/LightningFlowComponents/issues/875

Megalithe:
I am encountering this issue in the lookupFSC and updating to the QuickLookup component did not resolve. As mentioned the filters/where clauses do not apply to child lookups by default. I have to clear and reselect the preceding value to force the lookup to apply.

alexed1:
Fixed in version 1.1 thanks to @AndyHaas https://unofficialsf.com/lookup/

Issue #878, opened 2021-08-24
Quick Lookup - Pop-down not collapsing


CultechPtyLtd:
Issue: 
When clicking anywhere not on the component, the pop-down for displaying records down not collapse.

1. Click the lookup search field
2. A pop-down will appear for recent records. 
3. Click anywhere off the component

### Expected behaviour
When clicking anywhere not on the component, the pop-down should collapse.

### Actual behaviour
The pop-down does not collapse, which gets in the way of other components you may be trying to use which are directly below.

AndyHaas:
@CultechPtyLtd 
I have made the changes to the component to close if the user clicks off the component.

alexed1:
Fixed in version 1.1 thanks to @AndyHaas https://unofficialsf.com/lookup/

Issue #875, opened 2021-08-24
Quick LookUp - Clearing Parent Value


CultechPtyLtd:
### Steps to reproduce

1. On Parent Lookup, search and select a record.
2. On Child Lookup,  search and select a record.
3. Clear the Parent record.
4. Clear the Child record. 
5. Searching on child record still filters via the previous Parents value instead of null

### Expected behaviour
The Parent record should clear its value after clearing out the lookinig and thus clears out the Child filtering value. My use case I'm using the Parent as a Filter for the Child lookup. Sometimes you want to add a filter and remove the filter while searching throught the Child lookip.

### Actual behaviour
This bug doesn't allow the user to clear out the filter for the Child record.

### Screenshots
<img width="485" alt="Showing All Work Orders - No Parent" src="https://user-images.githubusercontent.com/73424220/130534788-5607a99a-c6f7-42e4-b742-5ad9a8bf6c1a.png">
^ Showing all child records, Parent filer not yet applied.
<img width="286" alt="Filtering WO, Parent Filter Added" src="https://user-images.githubusercontent.com/73424220/130534809-76a66122-c02e-4f81-93bf-f6e0ae0b93c7.png">
^ Filtering Child records, filtered by Parent record.
<img width="333" alt="Parrent Cleared, still filtering via previous parent value" src="https://user-images.githubusercontent.com/73424220/130534825-5de7a0fc-d208-4d8d-9b88-c7928675c4c0.png">
^ Clearing Parent record, Child still filtering by previous Parent value instead of null

alexed1:
Fixed in version 1.1 thanks to @AndyHaas https://unofficialsf.com/lookup/

PR #1221, opened 2022-09-16
added an event listener to listen if the user clicks outside the comp…

AndyHaas:
Adds an event listener to the component and if the menu is open and the user clicks out of the component the menu will close.

If parent lookup is cleared the child component(s) will be cleared along with their filters

Resolves Following Issue:
#878 
#875 
#962


alexed1 (approved):

alexed1:
@AndyHaas What I really love about these PR's is that you've solved a bunch of problems from different parts of the repo, and addressed other people's asks. There have been collaborators who built and maintain wonderful extensions, but we've always lacked a community of fixers and improvers. Just want to express my appreciation. 

If you want to share your email, I'll invite you to the private Slack workspace for Flow extension developers...it's a good way to converse.

PR #1227, opened 2022-10-03
Added FieldType of Time to allow mapping of time fields

AndyHaas:
Added the Time fieldType to the map collections. This was a missing type within the MapCollections. 

Screen Shot 2022-10-02 at 21 50 49

### DIFF
--- a/flow_action_components/CollectionProcessors/force-app/main/default/classes/MapCollection.cls
+++ b/flow_action_components/CollectionProcessors/force-app/main/default/classes/MapCollection.cls
@@ -75,6 +75,11 @@ global with sharing class MapCollection {
             }
         } else if (fieldType == 'JSON') {
             obj.put(fieldName, JSON.serialize(fieldValue));
+            
+        } else if (fieldType == 'Time')
+        {
+            string[] timeParts = fieldValue.split(':');
+            obj.put(fieldName, Time.newInstance(Integer.valueOf(timeParts[0]), Integer.valueOf(timeparts[1]), 0, 0));
         } else {
             obj.put(fieldName, fieldValue);
         }


AndyHaas:
@alexed1 hold off on this PR. As flow engine itself has issues with time coming from another record.

AndyHaas:
@alexed1 this is ready to go. The input has to be a type string. If the input type is not a string you will get the following error. 
Value of class com.force.swag.soap.Time is not a ValueType: STRING

You can not use an assignment to set a TEXT variable from a time field. You will need to create a formula field and use TEXT(Time__c) to convert it to a string. But this also only works on screen flows and if there is a screen for the formula to calculate. 

Overall this is an issue with Flow Engine as it doesn't know how to handle Time type and convert it to the appropriate type.

alexed1 (approved):

alexed1:
This line perplexes me: "But this also only works on screen flows and if there is a screen for the formula to calculate."
Can you expand on that? You're saying that if I create a Formula variable in my flow that attempts to reference a time field, it will work if it's a screen flow? How does the screen play a role?

alexed1:
also, can you express an end ot end use case for this? i.e an example we could put into the documentation?

AndyHaas:
@alexed1 
Sorry missed your comments on this one.

The action can be used for any flow type. I have tested further and saw that this would work for all other flow types. The only thing is that you need to convert the time variable to a string. 

Example:
The Account object has a time field, and you want to use that time field on every case created. You can get the time variable from the Account and convert that to a string by using a formula, TEXT(Account.TIME). Then use that formula variable to assign that to the case. 

You can not use the variable straight from the Account.

Issue #1223, opened 2022-09-19
*Lightning Input FSC* - _(Value)_ - Value can NOT be a variable or formula, only hard coded


JodieM:
### Steps to reproduce

1. Add Lightning Input FSC. 
2. Use Standard Settings as per documentation 

type: "number"
label: "Enter a decimal value"
step: "0.001"

3. Enter Value as 1234.01
4. Run
5. Click Previous
6. No Error. 

Now make only one change. 
3. Enter Value as a Number Variable with the same default value {!vNumber}
4. Run
5. Click Previous
6. Error: Flow encountered an error when processing and converting between data types. Please check the flow and ensure all data types are matched correctly

### Expected behaviour
Coming back to the screen should show the value in the variable again, not an error. 

### Actual behaviour
Error: Flow encountered an error when processing and converting between data types. Please check the flow and ensure all data types are matched correctly

### Screenshots
Here's a video https://www.loom.com/share/6c7772afbc214071b7e3d19f956b6bc8

### Debug Logs
If applicable, please insert debug logs here
<details>
<summary>Debug Log Information</summary>

```
55.0 APEX_CODE,FINEST;APEX_PROFILING,FINEST;CALLOUT,FINEST;DB,FINEST;NBA,INFO;SYSTEM,FINE;VALIDATION,INFO;VISUALFORCE,FINER;WAVE,FINEST;WORKFLOW,FINER
19:53:58.0 (177136)|USER_INFO|[EXTERNAL]|0056F000006aVRm|redacted|(GMT+10:00) Australian Eastern Standard Time (Australia/Sydney)|GMT+10:00
19:53:58.0 (45603121)|FLOW_VALUE_ASSIGNMENT|91204136eb267acb16bc71ebf0e618354e8496a-7e1e|error.errorId|358475656-319734 (-196161961)
19:53:58.0 (45664258)|FLOW_VALUE_ASSIGNMENT|91204136eb267acb16bc71ebf0e618354e8496a-7e1e|error.type|FLOW_INTERVIEW_HANDLED_ERROR
19:53:58.0 (45677191)|FLOW_VALUE_ASSIGNMENT|91204136eb267acb16bc71ebf0e618354e8496a-7e1e|error.message|Flow encountered an error when processing and converting between data types. Please check the flow and ensure all data types are matched correctly.
[...]

PR #1228, opened 2022-10-03
removed default value as this would cause an issue when a user clicks…

AndyHaas:
… previous button.

We have to remove the default value because when using a variable to define the value field, Flow doesn't understand what that data type is only when a user selects the previous button on a screen. On initial load, the screen and component load fine, but after navigating off that screen and navigating back to the screen, the error populate. 

This resolves issue
#1223


alexed1 (approved):

alexed1:
new version 1.0 package at https://unofficialsf.com/lightninginput/

PR #1232, opened 2022-10-07
Update fsc_flowButtonBar.js to add validation

Garywoo:
Added a validation handler so that users cannot navigate to the next flow screen without a button selection when the component is set to be required.


alexed1:
Thanks for the fixup, and welcome to the project!

alexed1 (approved):

jdayment:
@Garywoo there is an issue with the validation when there are multiple button bars on one screen (initial look appears it's related with session storage and using value to throw the error) we ran into something similar with text area plus and had to move our validation to separate js file to hand storage issues cc @alexed1

PR #1240, opened 2022-10-17
Update customHeaderCpe.html

dignacioconde:
Thanks to @vmandrespascual we have detected an error. We have deleted the conflictive vars in lightning-radio-group  It would be needed to add an alternative for the helper but with this change now it works in Winter 23 + orgs.


vmandrespascual (approved):
The change is correct, since according to the documentation of this component these 2 input parameters are not available.

[Salesforce lightning-radio-group DOC] (https://developer.salesforce.com/docs/component-library/bundle/lightning-radio-group/specification)

alexed1 (approved):

PR #1252, opened 2022-11-01
Invocable approval helper

Henk3000:
I wanted to use the flow action as shown on this article:
https://unofficialsf.com/use-flow-to-approve-reject-and-cancel-approval-processes/

As I now see in the comments the suggested code does not work, I investigated further and saw room for improvement.
I created 2 classes to be able to use the invocable approval actions
I removed the original flow example as it wasn't even calling the invocable action that was created in the article.

Other than that, it still works as described in the article


alexed1:
I realize now that Tamar also provide a pull request with fixes, at https://github.com/alexed1/LightningFlowComponents/pull/1281. 
I went ahead and merged her PR, but I'd still like to include your improvements if we can do it without changing the name of the action.

alexed1:
Hi, checking back again. Let me know if you'd like to discuss this PR.

Issue #1249, opened 2022-10-26
QuickChoice - Display Changes


ericrsmith35:
@AndyHaas THis was posted recently on the unofficialsf Trailblazer site. (https://trailhead.salesforce.com/trailblazer-community/feed/0D54S00000JePai)

![image](https://user-images.githubusercontent.com/24724447/198028241-09ad025a-608a-4066-a146-3144bc7c82b1.png)

It looks like the default display has changed.  Can this be corrected, or can you add an attribute setting that defaults to the original display mode and when set, implements your new standard height options?

AndyHaas:
This is a bug. I will take a look at this sometime today. The button size should still be an option on the setup of the component.

bookerg:
HI @AndyHaas 
Any update on this?  
Thanks

AndyHaas:
@bookerg,
Has the latest update fixed your issue?

bookerg:
> @bookerg, Has the latest update fixed your issue?

Hi Andy,

Just tested the latest release - and yes your changes have sorted this issue out.
Thanks

PR #1260, opened 2022-11-08
Reverting the cardSize section, but also keeping the changes by addin…

AndyHaas:
…g a checkbox on the CPE to handle if the boxes are desired to be the same height as the other column or not. If unchecked the cards will be independent, else the cards will be the same height regardless of content

#1249


AndyHaas:
Screenshot 2022-11-08 at 01 11 55

AndyHaas:
Screenshot 2022-11-08 at 01 58 50


Screenshot 2022-11-08 at 01 55 11

alexed1 (approved):

PR #1263, opened 2022-11-11
Update fsc_flowButtonBar.js

jdayment:
Comments out validate that was causing issues when multiple flow button bars are on the same flow screen. This was introduced in this commit e9bbfd9f09e973efb7e6fab4388a7773e11b4cce

Also adding in reactive screen support for Flow Flex Card


alexed1:
Thanks, @jdayment . @Garywoo are you aware of this issue? Do you want to take a look at how we can incorporate your validation without the multiple button bar problem? I'd like to keep the validation in the component....

Garywoo:
I was not aware of this issue. I'm not sure I have the skill to resolve this issue though. I can give it a shot when I have some free time though.

jdayment:
@Garywoo We handled validations for Text Area Plus by offloading our validations into a supporting .js file that sets and resets a key-value pair when multiple versions of the component are on the same screen and it's working well if you want to take a look at that.

fromsteinsfdc:
Is there any documentation or repro steps of the error?

alexed1 (approved):

alexed1:
I'm going to merge this revert to get the component back to full working state, but I'd love to reintegrate the validation piece once we solve for the multiple versions issue.

Garywoo:
I figured out the issue @jdayment reported here and resolved in my new PR #1302. Sorry about that!  
@alexed1 This reintegrates the validation behaviour, now that it's been fixed.

PR #1266, opened 2022-11-15
Knowledge tree continuation and end functions

tee-oh:
Added three new fields to Diagnostic Node object: Continuation_Node__c (boolean), End_Node__c (boolean), and Continuation Diagnostic Node (lookup to DiagnosticNode__c).

Added validation rule, if DiagnosticNode__c.Continuation_Node__c is true, DiagnosticNode__c.ContinuationDiagnosticNode__c lookup is required (and vice-versa).

Updated flow:
- Modified "Get Records:Load Current Node" to pull in the three new fields (Continuation_Node__c, End_Node__c, and ContinuationDiagnosticNode__c) and store them in {!curRecord} variable.
- Added two outcomes to "Decision:Does Outcome Have Target Node". "Yes, Has Target Continuation Node", condition Continuation_Node__c = true, and "Yes, Has Target End Node", condition End_Node__c = true. Added second condition to the existing outcome "Yes, Has Target Node", condition End_Node__c = false.
- Added assignment node "Set Next Node (From Continuation Node)", curDiagnosticNodeId = ContinuationDiagnosticNode__c to support continuation without selection functionality.
- Adding new screen "Display Finish Screen" to support end node functionality.

### DIFF
--- a/flow_apps/KnowledgeTree/force-app/main/default/flows/Twenty_Questions_Troubleshooter.flow-meta.xml
+++ b/flow_apps/KnowledgeTree/force-app/main/default/flows/Twenty_Questions_Troubleshooter.flow-meta.xml
@@ -38,7 +38,7 @@
         <description>If you reach a leaf node, it might not have a target node selected</description>
         <name>Does_Outcome_Have_a_Target_Node</name>
         <label>Does Outcome Have a Target Node?</label>
-        <locationX>1239</locationX>
+        <locationX>1242</locationX>
         <locationY>318</locationY>
         <defaultConnector>
             <targetReference>No_Node_Found</targetReference>
@@ -50,6 +50,13 @@
             <conditions>
                 <leftValueReference>outcomeChoices</leftValueReference>
                 <operator>IsNull</operator>
+                <rightValue>
+                    <booleanValue>false</booleanValue>
+                </rightValue>
+            </conditions>
+            <conditions>
+                <leftValueReference>curRecord.End_Node__c</leftValueReference>
+                <operator>EqualTo</operator>
                 <rightValue>
                     <booleanValue>false</booleanValue>
                 </rightValue>
--- a/flow_apps/KnowledgeTree/force-app/main/default/flows/Twenty_Questions_Troubleshooter.flow-meta.xml
+++ b/flow_apps/KnowledgeTree/force-app/main/default/flows/Twenty_Questions_Troubleshooter.flow-meta.xml
@@ -1,11 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Flow xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>49.0</apiVersion>
     <assignments>
         <description>Take the value of the radio button selected. That is the id of the next node to transition to.</description>
-        <name>Set_Next_Node</name>
-        <label>Set Next Node</label>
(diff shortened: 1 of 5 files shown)


alexed1:
I'm finally circling back to this...thanks so much for contributing this enhancement, Teo. I find myself frequently talking about this  'Thousands of Screens' pattern and am glad you've found it useful.

Both for me to better understand what you added here and for the benefit of future users, can you draft a Google Doc that explains a little more about what a Continuation Node is and what the use cases are for the new things you've added here?  We'll use that doc in a post about this enhancement.

tee-oh:
Hi @alexed1, yes I can. I can ping you on this PR with a link to that when I have it drafted up. Likely this weekend or so.

tee-oh:
Hello @alexed1, I was finally able to sit down and write out a google doc of what you requested. Could you let me know if you can access this and if the content works for you?

[https://docs.google.com/document/d/1TA5HHz8DXI2olcFNC93Sw0t3ggxfcwOetuAOK-g1AkE/edit?usp=sharing]

alexed1:
Fantastic write up. I'll merge the pr and get this published

alexed1 (approved):
This is a great enhancement, Teo!

alexed1:
I've always had trouble packaging this app because it reference Knowledge. I'm going to take some time to look into that before I publish your article, and see if I can't get it packaged.

tee-oh:
@alexed1 awesome! thanks! I'm glad I could contribute something meaningful!

alexed1:
I've learned that Knowledge is simply not packageable. The flow that's in the repo contains the Knowledge option and so has references to Knowledge. That makes that flow not packageable. Reading the original post, I did package up a version of the flow that didn't have knowledge. I can't find the xml source to that original flow, though.  What I think makes the most sense here is to create a scratch org, turn on knowledge, deploy the flow and the rest of the package to that org, recreate a version of the flow that doesn't have the knowledge option. save that to the project and then build a version that includes your new paths but doesn't include knowledge and can be packaged. How does that sound?

tee-oh:
@alexed1 Ok, that makes sense. Looking at the troubleshooter now and I can see the following areas that that need to be modified to avoid knowledge references:

In the flow file:
- remove the Use_Knowledge_Article_Content__c and Knowledge__c fields from the stored fields of the initial DiagnosticNode Get Records query.
- remove an Outcome from a Decision that evaluates the data source of the flow screen (diagnostic node or knowledge article).
- remove the Get Records knowledge query that Outcome points to.
- remove the knowledge-specific flow display screen that Get Records points to. 
image
- I think it might be worthwhile to retain the Decision that would have previously evaluated the data source of the flow screen (diagnostic node OR knowledge article) even though it really only has one "data source" for now (the node content field on the diagnostic node record) and use its Description area to give a clue to end users how they COULD incorporate knowledge into the flow by just adding a new outcome here and building out the pathway we are essentially removing.
image
image


Also, I can see in the layouts folder of the project, that the current DiagnosticNode layout displays a Knowledge__c lookup field (plus the Use_Knowledge_Article_Content__c checkbox field) and that there are three knowledge layouts included in this folder as well. Additionally, the objects folder also contains the knowledge object. I'm guessing these need to be removed also to enable packaging, is that right? Is that something you'll do or should I also pull them out?

For now, I'll create a new PR with the updated flow file that doesn't reference knowledge at all (other than in text in a description field like I suggested higher).

PR #1278, opened 2022-11-23
File Upload Improved CPE add and Flow Banner Update

jdayment:
Includes addition of a CPE for File Upload Improved and updates to Flow Banner to support rich text


alexed1 (approved):

alexed1:
https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G0000043xrTQAQ 3.0.16

sweet improvement, Josh. This File Upload has really become quite the showpiece. Would you be able to create a blog post describing the new configuration experience?

jdayment:
Yes I will

AndyHaas:
@jdayment 
Running into issues with the CPE I am not able to select any variables
Screenshot 2023-01-04 at 11 14 02

 or use the {!$Flow.InterviewGuid}
Screenshot 2023-01-04 at 11 17 09

jdayment:
@AndyHaas I am not able to duplicate this tried in a couple of orgs for the flow interview guid you can't reference it directly in fsc_flow_combobox but if you create a string variable and set the default to the interview guid it works you can technically set that to any unique string but hopefully will have this be an autonumber in the next release and this won't even be part of the cpe.

AndyHaas:
@jdayment 
Makes sense I would udpate the documentation then to inform users how to set this value because out of the box it doesn't make sense. Once I created the text field I was able to use the flow guid id. 

Thanks for the information.

PR #1293, opened 2022-12-05
New LWC Customizable Lookup Component

AndyHaas:
New LWC version of the Lookup component with some additional enhancements. Documentation to come.


AndyHaas:
Documentation:
https://docs.google.com/document/d/1-yW-7RBIMxDikP4eu3C-zE4zJ97mMSLGHkB1-sTYcuE/edit?usp=sharing

AndyHaas:
@alexed1 keep this open until we can complete several feats and enhancements for this component.

alexed1:
awesome. can you share the google doc?

alexed1:
There's a key structural question here and that's the decision to include this new comboBox as a separate, discrete screen component.  I recall a little conversation about that but not the result. The implication here is that someone (a developer? an admin? ) would be interested in using combobox separately. What's the use case? If we're going to make this an independent new screen component, we need documentation that explains what it adds to the community that isn't already available. I don't want to sound negative. This is actually pretty intriguing. But it's a big deal so we need clarity on it.

AndyHaas:
@alexed1 I will move the FSC combo box to this component and the FSC radio buttons to get the component out. Unless you think this would be a bad idea when we move this out of this component.

AndyHaas:
New PR #1350

PR #1298, opened 2022-12-10
Fix error message from being cut off

AndyHaas:
Added a top margin to the custom error message the combo box had
Added a bottom margin to the OOB input error message
Added a validity checker anytime the input has changed. this.hasError was never being triggered and be set true or false


AndyHaas:
Screenshot 2022-12-10 at 16 20 15

alexed1 (approved):
Love them small tweak improvements to existing extensions.

PR #1300, opened 2022-12-12
Add Frameborder to flowIframe and fix for FileUploadImproved

jdayment:
Adds frameborder attribute and property


alexed1 (approved):
Thanks for the enhancements, Josh!

PR #1323, opened 2023-01-12
New timeZone component

AndyHaas:
New APEX action that calls the Google TimeZone API to get the timezone from a lat and long on an address field.


alexed1 (approved):

alexed1:
I'm going to move this to a new GoogleActions project.

PR #1329, opened 2023-01-17
flowDatetimeMethods moved from Carls repo to USF collection

AndyHaas:


alexed1:
in a typical scenario i think using submodules would make a lot of sense. but one of the things I'm prioritizing is a reduction in pointers to code from other repos, in the interest of security. So can you redo this as a pure copy of code, Andy?

AndyHaas:
Closing this as I am moving to another PR

PR #1333, opened 2023-01-26
Andretawa open ai

andretawa:
OpenAI Connect invocable action for flow.


alexed1 (commented):
This looks great. Do you have a Google Doc we could use as an accompanying post? Also it would be great to have a screencast showing it in use.

andretawa:
> This looks great. Do you have a Google Doc we could use as an accompanying post? Also it would be great to have a screencast showing it in use.

Hi Alex, thanks! Google docs is coming latest tomorrow. A link to the screencast is in your email. It was a bit to big for github :)

andretawa:
Blog post is finished as well: https://docs.google.com/document/d/1mvOkOh_V3kJ5Q4c9-mwisOtQiO_nJdshOhvqui0Mqtc/edit?usp=sharing

andretawa (commented):
All fine.

alexed1 (approved):

alexed1:
@andretawa , packaging is failing because the flow contains a reference to the custom object GPTConfiguration__c . We need to either add that custom object to the project (meaning it would get created on all orgs where this flow is installed) or remove the flow from the package or modify the flow to not use a custom object.

I've run into this before, and taken different paths depending on the situation. In some cases, it just wasn't feasible to provide a sample flow because it would require too much additional metadata to be generated on the org. In other cases, I've concluded that users will probably install the package to a sandbox and then potentially deploy just the action class to production, so it's not that big a deal. 

What's your thinking on this?

andretawa:
@alexed1 created another pull request. Should be working now. Otherwise, just delete the flow.

PR #1337, opened 2023-02-02
Fsc Combo Box

AndyHaas:
This is used within the new FSC lookup component. Documentation to come. 

```    
<c-fsc_combobox options={records} label={label} icon-size="small" public-class={publicClass} public-style={publicStyle} values={values}
        custom-search-handler={handleSearchChange} required={required} allow-multiselect={allowMultiselect} disabled={disabled}
        is-loading={isLoading} exclude-sublabel-in-filter={excludeSublabelInFilter} placeholder={placeholder} field-level-help={fieldLevelHelp}
        include-value-in-filter={includeValueInFilter} onchange={handleComboboxChange} right-icon={rightIconName} left-icon={leftIconName}
        oncustomaction={handleCustomAction} debounce-delay={debounceDelay} no-match-string={noMatchString} message-when-value-missing={messageWhenValueMissing} 
        id="lookup" class={componentName}>
    </c-fsc_combobox>
```

### Inputs ###
Options Array of Objects of Label, Value Pair [{"Label","Value"}]
Label - String - The label for the component
icon-size - String - The size of the label next to the label 
public-class
public-style
values
required - Boolean - will require the field or not
allow-multiselect - Boolean - will enable/diable multiselect
disabled - Boolean - will disable field manipulation or not
is-loading - Boolean - will show loading icon
exclude-sublabel-in-filter
placeholder
field-level-help
include-value-in-filter
right-icon
left-icon
debounce-delay
no-match-string
message-when-value-missing

### Outputs ###
custom-search-handler - Returns string of text the user is searching
onchange - Value in the event.detail.values. Fires when a selection is made
oncustomaction - New Record Action


alexed1:
Great! Can you add here:
1) what are the 1 or 2 top use cases for this component
2) how it is similar/different to the Flow Combobox component
3) where it is being used now

alexed1:
can you elaborate on the comment about dependencies? Which component has the depenency on which components?

AndyHaas:
@alexed1 
I'm sorry for not getting back to you sooner. I needed to look at the fsc_flowCombobox code. In short, this component would be used within CPE elements where, whereas this component would be used in the LWC context. This one has more flexibility to allow further elements to be configured. 

Now that I say this. The question comes; can they be merged? Maybe. We can combine these two with enough work and validations and have one component. 

If we want we can build this into the lookup component and when the original component gets updated to meet the needs that lookup needs we can migrate those changes over.

AndyHaas:
> can you elaborate on the comment about dependencies? Which component has dependent on which components?

This was about the lookup component itself.
    "MC_Combobox": "04t5e0000012DeJAAU", ( This Component )
    "SOQLBuildAndExecute@1.3.1-0": "04tf4000004PtguAAC",
    "FlowActionsBasePack@3.6.0-0": "04t8b000000mKnoAAE",
    "FlowScreenComponentsBasePack@3.0.6-0": "04t5G000003rUvVQAU",
    "fsc_FlowRadioGroup": "04t5e000000NqxCAAS"

alexed1:
Thanks, Andy. As I mentioned the other day, it's useful to reserve the fsc prefix for stuff in the base components. It makes it easier to immediately spot problems. So we'd want to rename this one too.

I'm happy to check it in but I want to include a little more documentation on what it does. As you point out the Flow Combobox in base components is designed for use in CPE's and focuses on provided mergefield picking.  check my understanding:
What you have here doesn't go into CPE's and isn't merge field aware. It's a general purpose combobox control great for providing typical typeahead search based on the data source it's attached to. 

What are the things that this component does that go beyond just using the base https://developer.salesforce.com/docs/component-library/bundle/lightning-combobox/documentation Explaining how this component pushes things forward is key to enabling others to use it.

Is there anything flow-specific about this component? Might I want to use it in an arbitrary LWC that has nothing to do with flow? or does it have flow awareness built into it in some way?

We should think of a name that expresses that extra capability like comboBoxEnhanced

AndyHaas:
Closing as I am going to embed this into the lookup component.

PR #1341, opened 2023-02-07
DetectAndLaunch Bug Fixes

AndyHaas:
changed how helper class was deciding to launch flow or not. Some flows got launched even tho they were not being set. Also moved Loaded to its own else if statment and made seperate class to call the flow to have unified code between loaded and changed events.


alexed1:
v2.2.1 https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G000004C6tfQAC

PR #1342, opened 2023-02-08
Fsc flow radio button group

AndyHaas:
This element is used for CPE development. Allows users to choose a cleaner way to choose between two values 

##Inputs                   
name = The name of the input field
label = The Label for the field
options = The options for the user to select. Usually, two values ( like true/false )
value = The current value of the field
type  = How to display the radio button. ( Button or Radio are valid options )
##Outputs
Action: onvaluechange
Value: event.detail.newValue ( this follows naming convention of other fsc components )

Screenshot 2023-02-07 181650


lightning_page_components/DetectAndLaunch/force-app/main/default/aura/detectAndLaunch/detectAndLaunchHelper.js
```diff
@@ -1,8 +1,28 @@
 ({
```
alexed1, line 1: is it intentional to have this Detect and Launch file in this pull request with the radio button stuff?

AndyHaas: Created new PR #1348

PR #1343, opened 2023-02-10
Update OpenAI_Connect_Template.flow-meta.xml

andretawa:
@alexed1 Removed the reference to the custom object. If it doesn't package still, just get rid of the flow then.


alexed1 (approved):

alexed1:
We're up at https://unofficialsf.com/tap-into-the-power-of-chatgpt-via-flow/ !
Thanks for your contribution Andre! Let us know if you come up with new ideas.

PR #1356, opened 2023-02-27
quickLookup - Output Display Value

RealEvanPonter:
I've added an output attribute that will allow for referencing the name of the record that was selected in the quickLookup component in downstream flow configurations.

### DIFF
--- a/flow_screen_components/quickLookup/force-app/main/default/aura/quickLookup/quickLookup.cmp
+++ b/flow_screen_components/quickLookup/force-app/main/default/aura/quickLookup/quickLookup.cmp
@@ -11,6 +11,7 @@
     <aura:attribute name="valueFieldName" type="String" default="Id" />
 	<aura:attribute name="label" type="String" />
 	<aura:attribute name="selectedValue" type="String" />
+	<aura:attribute name="displayedValue" type="String" />
 	<aura:attribute name="filterFieldName" type="String" />
 	<aura:attribute name="filterFieldValue" type="String" />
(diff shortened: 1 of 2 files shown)


alexed1:
Thanks, Evan! Can you provide a simple example of this in use in a Google Doc? Doesn't have to be fancy....

alexed1:
Hi, @RealEvanPonter . Checking in again on this. I'd love to get a little documentation or a screen shot...

RealEvanPonter:
Yes - I have a Quick LookUp component on a Flow screen that allows for selecting an Account record. The options needed to be filtered down using a WHERE clause to limit the results based on a record that was selected earlier in the flow. The record Id output is used during the creation of a new Opportunity record to relate that record to the Account. I wanted to display a confirmation page at the end of the flow that shows all the values specified during the flow. Displaying the Account Id would not be the best user experience here - I wanted to show the name of the record instead. Since the Quick Lookup component already had all the information, I wanted to be able to reference it in the flow without having to add a Get Records element to retrieve it.
image

Here is the confirmation page that uses the Output Display Value (Account.displayedValue) from the Quick LookUp component:
image

alexed1 (approved):

alexed1 (approved):

alexed1:
Implemented in version 1.2 https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G000003zgYwQAI along with https://github.com/alexed1/LightningFlowComponents/pull/1426#pullrequestreview-1514581101

PR #1384, opened 2023-04-05
feat: enable and disable rich text formats

clifford-fra:
Hello,

I'm using the richTextInput component to create and edit ContentNotes in Flow. The feature (format) disparity between the Salesforce Note Editor and this component caused some bugs, because some formats are not shown in the Salesforce Editor and vice versa. We have also encountered errors inside `<lightning-input-rich-text>` when creating links with the Link-Button.

For these reasons, I have updated the component, so that it uses the `disable-categories` attribute to remove button groups from the standard toolbar. Additionally, I also added an `enabledFormats` attribute in case someone wants to specify the toolbar feature set exactly.

- Both new attributes are optional
- I updated the documentation
- The new version is downward compatible to the existing version
- I have tested the new version with the following cases
  - enabledFormats is empty and disabledCategories is empty
  - enabledFormats is empty and disabledCategories is not empty
  - enabledFormats is not empty and disabledCategories is empty
  - enabledFormats is not empty and disabledCategories is not empty

So, basically, I implemented the formats and disabled-categories from here https://developer.salesforce.com/docs/component-library/bundle/lightning-input-rich-text/documentation while being downward compatible.

Best regards
Clifford

### DIFF
--- a/flow_screen_components/richTextInputFSC/force-app/main/default/lwc/inputRichTextFSC/inputRichTextFSC.js
+++ b/flow_screen_components/richTextInputFSC/force-app/main/default/lwc/inputRichTextFSC/inputRichTextFSC.js
@@ -1,5 +1,6 @@
 // 06/25/21    Eric Smith       Added a Required attribute and increased the bottom margin to better match standard components
 // 07/02/21    Eric Smith       Added a Required * to the Input Label if Required is True
+// 04/04/23    Clifford Beul    Added disabled categories and optional overwrite of formats
 
 import { LightningElement, api, track } from 'lwc';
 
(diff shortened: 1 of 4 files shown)


flow_screen_components/richTextInputFSC/force-app/main/default/lwc/inputRichTextFSC/inputRichTextFSC.js
```diff
@@ -15,6 +16,8 @@ export default class inputRichTextFSC_LWC extends LightningElement {
     @api label;
     @api characterLimit;
     @api required = false;
+    @api disabledCategories = '';
```
clifford-fra, line 19: The `<lightning-input-rich-text>` component will throw an indexOf exception, if disabledCategories is not initialized with an empty string.

alexed1 (approved):
Thanks for this great enhancement, Clifford!

PR #1387, opened 2023-04-14
Flow Auto Navigate | Added the ability to Show Loader Icon

AndyHaas:
We have added the ability to show a loader in lieu of a timer.

This is what the user will see for the duration of the timer.
Screenshot 2023-04-13 at 9 48 35 PM


alexed1:
Thanks, Andy. Can you provide a google doc with a blurb of text and images that show how it looks like when this new attribute is false vs true? It doesn't have to be more than a couple of sentences and images but that will make it easy for me to add the content to the existing blog post when this gets built into the new package.

AndyHaas:
@alexed1 here is the documentation, the change was quite simple to be honest

https://docs.google.com/document/d/10iTkEUyNuJ0AAzSg-P_c1xMVtqvcyVHuYvlgZJ3_W1Q/edit?usp=sharing

alexed1 (approved):

PR #1394, opened 2023-04-28
CollectionCalculator | Update | Added new features

AndyHaas:
There are some similar components in USF, but it doesn't have them all combined in one component, so we can close this out if this is considered a duplicate. But with this APEX Action, it has the main math function that one would use in FLOW that is currently not available. 

-- Added new SObject Aggregate Calculator to use the following actions SUM, COUNT, AVERAGE, MAX, MIN


alexed1:
Thanks for this contribution, Andy! It looks like the closest match is this[ CollectionCalculate action](https://unofficialsf.com/sum-multiply-and-average-fields-in-a-collection-with-collectioncalculate/) that's in the collection processor package.It supports Sum, Count and Average. Given that, we should probably make any changes in this area as enhancements to that existing action rather than introduce a different way to do them. Max and Min would be great adds. Thoughts?

AndyHaas:
@ericrsmith35 here are the changes to the CollectionsCalculate action

flow_action_components/CollectionProcessors/force-app/main/default/classes/AddOrInsertRecord.cls
```diff
@@ -1,5 +1,5 @@
 global with sharing class AddOrInsertRecord {
-    @InvocableMethod(label='Add or Insert Record [USF Collection Processor]' category='Util' iconName='resource:CollectionProcessorsSVG:colproc')
```
alexed1, line 2: It looks like this PR doesn't have the recent PR with thhe category and icon names. @AndyHaas , can you do a git pull and then update your PR so it's no longer trying to remove these lines?

AndyHaas: My last push fixed this issue.

AndyHaas:
Will redo PR as my code was completely out of date when I made my changes

PR #1395, opened 2023-04-30
DETECT AND LAUNCH | BUG FIX | Editing Page

AndyHaas:
One issue with this component arose during editing of the record page, as the DetectAndLaunch would launch based on the configuration. The current PR will examine the URL and prevent the DetectAndLaunch from launching if the page is the flexipageEditor.

I have conducted tests with both modal and modeless and it is working as intended.

Commit Comments
fixed issue when editing a record page the detect and launch would launch when a loaded URL was added


alexed1 (approved):

alexed1:
Thanks. good fix.

PR #1405, opened 2023-05-15
Text Area Plus and Flow Flex Card Updates

jdayment:
Text Area Plus
- Adds a Text Area Height property allowing for the text input area to have a minimum height

Flow Flex Card
- Fixes an issue where the actions modal was not rendering all flow components properly
- Updates "list" option for actions to use vertical navigation for better accessibility
- Adds label option for actions when in "list" mode
- Adds fsc_modalFlow for better performance and better compatibility with screen readers when launching sub-flow


jdayment:
@alexed1 I updated this PR to include some bug fixes for Flow Flex Cards I ended up creating a new component that handles the modal for the sub-flows previously I was using slds-modal and moved to lightning-modal something broke somewhere using slds-modal and lightning-flow and this was the solution don't know why

alexed1 (approved):

PR #1407, opened 2023-05-21
Datatable 4.0.13

ericrsmith35:
@alexed1 This is the source for my latest release (v4.1.0).  Some of the major enhancements include:

- Search Bar
- Flex Column Widths
- Use new LWC method to load the Column Wizard flow instead of the old Aura component


alexed1 (approved):

ericrsmith35:
I've removed the conflicting file

PR #1414, opened 2023-06-07
Update ResolveApprovalRequest.cls

Frederik-Witte:
The comment is wrong, the action is either "Approve" or "Reject" not "Approved" or "Rejected"


alexed1 (approved):
Thanks for the fix!

Issue #1425, opened 2023-06-26
*flow_action_components* - _SendBetterEmail _ - Attachment version compares as string


starkguerraj:
### Steps to reproduce

1. Create a file with more than 10 ContentVersions.
2. Send Email with component.

### Expected behaviour
Most recent ContentVersion should be sent.

### Actual behaviour
Comparison of versions < is done on the string typed VersionNumber resulting in ContentVersions beyond version 9 not being recognized.

### Screenshots
![image](https://github.com/alexed1/LightningFlowComponents/assets/72459978/5a038bbf-e244-4d0e-9c76-24e7d145d167)

PR #1427, opened 2023-06-26
Compare ContentVersions VersionNumbers as Numbers #1425 

mykah89:
Compare as decimals if the versions are numeric. #1425


alexed1 (approved):
Thanks for the fix, mykah!

PR #1428, opened 2023-06-27
Add WITH SECURITY_ENFORCED to queries

SFDX-Sam:
Adding WITH SECURITY_ENFORCED to queries to validate permissions


alexed1:
Thanks for this contribution...it's always great to have people keeping an eye on security. Can you tell me what adding this does to the query? My main question mark for something like this is: is there a risk we're going to break existing users and installations?

PR #1431, opened 2023-07-06
Fix null value in Flow Flex Card

jdayment:
@alexed1 this one is a hotfix it was impacting values not being returned in single select in Flow Flex Cards so will need a new basepack I had a line commented out that wasn't supposed to be.


alexed1 (approved):

alexed1:
incorporated into version 3.2.2

PR #1438, opened 2023-07-22
Qc reactive default

ericrsmith35:
@alexed1 Please create a new FlowScreenComponentsBasePack version.  This is an update to QuickChoice to support a default value being available for a reactive output and updates the choiceValues, choiceLabels and choiceIcons to be reactive input attributes.


alexed1:
Eric it looks like this json has a conflict (see above)...can you first resolve that? 
image

ericrsmith35:
@alexed1 Resolved

PR #1459, opened 2023-09-20
Update ConvertDateToDatetimeFlowAction.cls

StreckerCM:
Moved response initialization inside the for loop to allow for proper bulkification when creating/updating multiple records in a flow


ericrsmith35 (approved):
Change look good.

ericrsmith35:
Thanks, I'll update the install package.

PR #1468, opened 2023-10-03
DualListBox - Added new param "Select All" to have a user select all of the availab…

spider-bkv:
Hi all,

Added new param "Select All" to have a user select all of the available options. Params "Min" and "Max" will be ignored.

I found it useful for others.

Thank you!

### DIFF
--- a/flow_screen_components/dualListBoxFSC/force-app/main/default/lwc/dualListBox2/dualListBox2.js
+++ b/flow_screen_components/dualListBoxFSC/force-app/main/default/lwc/dualListBox2/dualListBox2.js
@@ -13,6 +13,11 @@ export default class dualListBoxFSC extends LightningElement {
     @api fieldLevelHelp;
     @api selectedLabel;
 
+    @api 
+    get selectAll() {
+        return (this.cb_selectAll == 'CB_TRUE') ? true : false;
+    }
+    @api cb_selectAll;
     @api min;
     @api max;
     @api disableReordering;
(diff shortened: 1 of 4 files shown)


ericrsmith35:
Are you looking to offer a Select All option to the user running the flow or are you looking to add a validation that requires the user to select every option?

alexed1:
Thanks for the enhancement, Kirill!

What's your favorite example use case where the user would be asked to select everything?

spider-bkv:
> Thanks for the enhancement, Kirill!
> 
> What's your favorite example use case where the user would be asked to select everything?

Thank you for the question @alexed1 !

So, the common use case description is "prioritization of a list of items" and when it's mandatory to use all of the items it's good to have the "Required to select all" parameter, instead of putting vars to Min and Max (btw, it's not possible now, also).

The most recent request I got was asking the user to prioritize a list of apartment properties (ie Floor Space, Light/Views, Closet Space, Unique Layout). After about a year in production, we've got good results from the users, it was clear to them what to do with the DualListBox moving items one-by-one from left to right in the highest-lowest order and having the ability to move them upside-down when they moved them to the right.

Thank you.

alexed1:
Got it... ok, I'd like to propose a few small tweaks:
1) move your checkbox from above the Min and Max to below. That will make it clearer that Min and Max are not dependent on the setting of the checkbox.
2) change the text to 'All items must be selected (use for sorting applications)'

alexed1:
Just checking in here, @spider-bkv . Do you think those tweaks would be possible?

alexed1:
Closing for now, but happy to reengage on this.

PR #1476, opened 2023-11-10
Update FieldPickerController to API 59

jdayment:
Updates to API 59 to support new core Non-Profit  cloud objects


alexed1 (approved):

alexed1:
should I repackage for this? Does it facilitate anything immediately?

ericrsmith35:
We have a few updates now for the FlowScreenComponentsBasePack that should
warrant a new version.

jdayment:
What Eric said 😅this was feedback from someone using the new Non-Profit
cloud some objects were missing

jdayment:
@alexed1 Can you also do a new version of FlowActionsBasePack this should be part of that package

PR #1478, opened 2023-11-17
allow full row search when empty cells are present

clev32:
Currently the row search will evaluate rows until it hits an empty cell in which case it breaks out of the loop without searching the rest of the cells in the row. By using a 'continue' statement instead of a 'break' statement the entire row will get evaluated even with the presence of empty cells.


ericrsmith35 (approved):
Thanks for the fix

alexed1:
Nice!

crouseamplify:
Thank you for this fix!

PR #1523, opened 2024-02-13
Update ManageMultiselectField.cls

gardnermt:
If the multipicklist has no values selected, then its value will be null. Running .split() on a NULL string variable will throw an error.

We need the code I added in line 14 to help prevent such errors so that this code can run in a record triggered flow or scheduled flow where some records in the execution may not have any values selected in the multipicklist.

Let me know if you have any questions.


alexed1:
added to version 1.0.3 https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G000004XZl7QAG

PR #1524, opened 2024-02-14
Datatable-Fix column value for apex defined objects, if source field is empty

spyros-michailidis:
Checks if the source field has a value and only then it populates value to the column. Without that check, if the source field is empty, the component populates the base url(MYDOMAIN) as label and active link to the column.


ericrsmith35 (commented):
Thanks for the fix.  I will include it in the next release.

spyros-michailidis:
> Thanks for the fix. I will include it in the next release.

You are welcome Eric and thank you for the prompt review.

PR #1528, opened 2024-03-19
use String.valueOf to set fieldValue variables instead of null method.

benkean:
When trying to compare a list of records that had a unique id that could be null, the use of the `.toString()` method causes  `EXCEPTION_THROWN [109]|System.NullPointerException: Attempt to de-reference a null object`. Changing this behavior to use `String.valueOf(toConvert)` avoids this exception.

I've added a test method to handle this error, as well.


alexed1:
Thank you!

PR #1529, opened 2024-03-22
Fix issue with External Objects date fields

philipnovak-pentair:
Salesforce has a weird behavior with External Objects where date fields are formatted as datetime fields (they include the time in the string representation). This was causing the values to be processed incorrectly here. `Date.parse()` was returning `NaN` and then `record[date]` was being overwritten with `null` as a result.  By checking for `isNaN` we can prevent this from happening.

See open issue from my colleague Ben: https://github.com/alexed1/LightningFlowComponents/issues/1124


ericrsmith35 (commented):
This looks ok.  I will include it in my next release.

alexed1:
Thanks for the fix, Philip!

PR #1538, opened 2024-04-29
Flow Flex Card Updates

jdayment:
Updates to allow for user to select the field in the header

Also includes updates from Brian Paul to dispatch changes directly to card from actions launched in modal really great addition

CPE updates include new banner help

Also includes component label in flow builder

Introducing Approval Checker

Invocable action that checks to see if a record is locked/unlocked and returns boolean. I have a post ready to go once it's packaged.


alexed1 (approved):

alexed1 (approved):

alexed1:
Thanks. For the approval actions, should I create a new package? i had a thought that perhaps we should merge this with the small existing package at https://unofficialsf.com/approval-processes-unlock-records-in-lightning/. Since we're shipping Lock and Unlock as standard actions, we should retire those particular ones. Let me know what you think the packaging plan should be going forward.

jdayment:
I think this should be a new package. I agree we should retire the other
one. I developed this with the core action in mind solely I have a post
ready to go covering the use case using the core action

PR #1542, opened 2024-05-28
Update fsc_quickChoiceFSC.js

bjvndrwalt:
Clear selected picklist values in dependent picklist when a value in the controlling picklist changes by:
1. Updating the elements priorControllingValue with previously selected controlling value.
2. Using the difference between controllingValue and priorControllingValue as a condition to set the element value = null


ericrsmith35:
Nice update.  Thanks.

BaileyAnn:
Would love to see this fix - any idea on when it will added to the QC in the flow base packs/action packs?

alexed1 (approved):

alexed1:
Here's version 3.3.4 of screen component base pack: https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5G000004fz7mQAA

PR #1543, opened 2024-05-29
Made single scan reactive

tschug:
Added the reactivity for single barcode scan.  Still unsure how to make continuous scan reactive, though that could be an undesirable situation.

This modification allowed me to present a lookup field as well as the scanner.  The scanner populated a formula variable SOQL which was used by data fetcher to then populate the lookup field's default value.

It allowed the user the ability to scan a barcode that wasn't a record Id, but another unique identified that data fetcher could then locate the correct record.


jdayment:
Nice update!

alexed1 (approved):

alexed1:
Thank you for this enhancement, tschug!

PR #1549, opened 2024-06-20
Object And Field picklist updates

StreckerCM:
Updated _field_ property to support multiple fields when _allowFieldMultiselect_ is true

Implemented _masterLabel_ property on the HTML template.  This property was captured but was not implemented on the template.

Removed the property _testproperty_ because it was not used anywhere and looked to be remnant of debugging.

Updated the _field_ property to allow multiple selected fields to be passed into the component when _allowFieldMultiselect_ is true.

There is a check to see if the fields in the property _field_ is available on the selected object.  This check existed in the loop that got all the objects field.  It did not use any of the variables from the loop.  If the pass field did not exist on the object the property would be cleared, and the check would only occur on the first loop.  However, if the field is part of the selected object this check would occur on each loop.  I moved this check outside of the loop prior to the start of the loop.


ericrsmith35 (commented):
Does leaving the API version at 46.0 mean that newer objects are unavailable to the component?

jdayment:
I don't think that should impact it the fields and objects should be
returned from the controller the API version of the controller should be
what the results are but I think we will want to update component APIs
going forward though based on the latest release notes I think API version
will start having and impact on lightning component updates.

StreckerCM:
Also I see another component _ObjectAndFieldPickerFSC_ that is outside of the _FlowScreenComponentsBasePack_ does that need to be in sync with the base pack version?

StreckerCM:
@jdayment and @ericrsmith35 I updated the API version to 60.0 on this component

alexed1:
StreckerCM, thanks for these enhancements! We don't get such careful scrutiny of the screen code very often, and I relly value it. 

I'm pretty sure that the component with FSC on the end is obsolete. All the modern components that go into the screen components base pack start with 'fsc_' so you can ignore the ObjectAndFieldPickerFSC.

@jdayment  can you code review this?

jdayment (approved):
I appreciate the descriptive comments everything looks good on my end.

philryan29:
> Removed the property _testproperty_ because it was not used anywhere and looked to be remnant of debugging.

I am installing from USF and received this error FYI:
fsc_pickObjectAndField3: lwc/fsc_pickObjectAndField3/fsc_pickObjectAndField3.js-meta.xml: You can’t remove the following public properties: testproperty, because the component is part of a managed package.

StreckerCM:
> > Removed the property _testproperty_ because it was not used anywhere and looked to be remnant of debugging.
> 
> I am installing from USF and received this error FYI: fsc_pickObjectAndField3: lwc/fsc_pickObjectAndField3/fsc_pickObjectAndField3.js-meta.xml: You can’t remove the following public properties: testproperty, because the component is part of a managed package.

I created issue #1587 for this.  I will add _testproperty_ back to the component to resolve the upgrade error.

PR #1569, opened 2024-08-01
inputRichTextFSC - remove html from character count in 

djtoohey:
Howdy, we found that it was confusing for users that the character count did not match what was being written.
The current character count for inputRichTextFSC was including HTML characters that were not visible to the user.
We introduced a regex that would remove anything between < and > from the character count. 

Only downside is that the field that the text area is linked to needed to be increased to handle for this change. 
For that reason, I'm not 100% sure if this is something that people have been wanting. But here it is 😄 

Maybe add a new attribute to the component to include the html characters in the character count or not?


djtoohey:
Update: have added in a new attribute to the component that determines if HTML characters are to be included in the word count.

ericrsmith35:
@djtoohey I like your enhancements but they shouldn't change the default behavior of the component for places where users have already implemented an earlier version.  Please change the attribute to something like **Exclude** HTML Characters from the Count.  This way, a default value of false will make the component behave the same way as it did before your changes.

djtoohey:
@ericrsmith35 good point, have made the changes necessary.

alexed1:
Thanks, Declan! This is a nice enhancement that increases accuracy.

alexed1 (approved):

PR #1580, opened 2024-09-05
fix syntax errors preventing compilation

tggagne:
probably the result of a bad merge


alexed1 (approved):
thank you for catching and fixing this!

PR #1598, opened 2024-10-23
Handle datatable not displaying updates reactively after column filtering

jaggimbert:
Found that while passing in records dynamically, that after filtering by a column, the table ceases to reflect updated input data. Adding in ```this.isUpdateTable = true``` has fixed the issue for me.

Use case here was a standard flow text field being used as a search bar, passing that term to a dynamic query in Data Fetcher, then sending those records from data fetcher to the table to update.


ericrsmith35:
I'm going to hold off on this change because I'm unable to reproduce the issue (v4.3.3)

PR #1600, opened 2024-10-29
Picklist values improved

Henk3000:
Changes:

1. Added checks for SObject Api Name and Field Api Name existence
2. if it does not exist provide valuable feedback for the admin (including a suggestion)
3. Suggestions made via Levenshtein Distance calculation
4. Test class include bulk testing (multiple fields)
5. Test class to include/test error handling


alexed1:
Very nice!

alexed1 (approved):

PR #1617, opened 2024-12-27
Combo box actions

ericrsmith35:
@alexed1 Happy Holidays!

This is the long awaited update for ComboBox to handle Action Buttons and Screen Actions.

Could you please process this Pull Request and create a new package for the FlowScreenComponentsBasePack.

I also updated the component to act more like the standard resource picker by allowing the user to click on the reference name instead of just the > character to drill down for additional options.

Since this doesn't yet support Filter and Transform outputs, I've added an option so CPE developers can allow a manual override for some inputs so the user can enter a {!reference} directly without causing an error.


ericrsmith35:
@alexed1 Please hold on this for now - I need to do some more testing.

ericrsmith35:
@alexed1 This is now good to go.  Please build a new FlowScreenComponentsBasePack

alexed1 (approved):

PR #1635, opened 2025-02-28
Enhanced picklist selection when overflow is not enabled

ChewinChaz:
This minor update allows easier picklist selection within datatables that are not allowed to overflow, especially when a small number of records are present.   I'm not a master coder by any means so I'm sure there are ways to clean it up a bit but the functionality works perfectly for my use case so I wanted to share in case you were interested.  Thanks for all the awesome things you do!

OverflowIssueFIXED


ericrsmith35:
Thank you for the details.  I will take a closer look at this.

PR #1637, opened 2025-03-19
add special case for KnowledgeArticleVersion in FindRecordsInCollection

KNLangston:
add special case in FindRecordsInCollection since KnowledgeArticleVersion is not available for Schema.getGlobalDescribe().get()

### DIFF
--- a/flow_action_components/GetPicklistValues/.vscode/settings.json
+++ b/flow_action_components/GetPicklistValues/.vscode/settings.json
@@ -4,5 +4,10 @@
     "**/node_modules": true,
     "**/bower_components": true,
     "**/.sfdx": true
+  },
+  "xml.preferences.showSchemaDocumentationType": "none",
+  "workbench.colorCustomizations": {
+    "statusBar.background": "#2f53a8",
+    "activityBar.background": "#2f53a8"
   }
 }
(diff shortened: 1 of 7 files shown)


ericrsmith35:
@KNLangston Do you know of any other objects this might apply to as well?

KNLangston:
@ericrsmith35 I was not able to find any documentation of what objects are unsupported and I've not run into this problem with any other objects before.  
However, I found a snippet of anonymous apex code to print objects which are accessible in the permissionsets but aren't included in the global describe here: [Objects not in getGlobalDescribe() (Scorecard, ScorecardMetric, ScorecardAssociation)](https://salesforce.stackexchange.com/questions/251711/objects-not-in-getglobaldescribe-scorecard-scorecardmetric-scorecardassocia).  Though, KnowledgeArticleVersion did not show up in the output and the objects described by the author (Scorecard, etc.) did print when I tried it today, which tells me that the list of unsupported objects is incomplete and will change over time.  
Perusing through the list, I did not spot any more object types that I would need in Flow, but this is subjective and will differ from person to person.  Do you want me to add more special cases for any of these other object types?
[object permissions minus getGlobalDescribe.csv](https://github.com/user-attachments/files/19348744/object.permissions.minus.getGlobalDescribe.csv)

ericrsmith35:
Yes please.  Add additional ones that you have found.

Or another option would be to do a Try/Catch the original way and add you
new option if that fails.

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Virus-free.www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

PR #1640, opened 2025-04-12
New Apex Action StringNormaliser

jockrlos685:
Converts a string with accented characters (diacritics) into their standard equivalent version (ASCII); eg. José > Jose
Includes additional methods such as:
- Removing special characters keeping alphanumeric only, eg. O'brian > Obrian
- Replacing special characters with empty spaces, eg. O'brian > O brian
- Returning using proper case (first letter capitalised), eg. joe doe > Joe Doe
- Removing all spaces
More info in the ReadMe field


ericrsmith35 (approved):
I will create a post for unofficialsf based on your Readme.  I'll add the link to this PR when ready.  I would appreciate any additions you could suggest for the writeup.  I will also repackage this and update the install links.

ericrsmith35:
https://unofficialsf.com/stringnormaliser-apex-action/

PR #1656, opened 2025-07-17
Edit fsc_flowButtonBar.css to enable css variable override

readeral:
At the moment, the cssString cannot influence the padding set on the lightning-button components because of the sibling selector.

This adjustment does a few things:
1. Changes the property to margin-inline-start: futureproofs for any changes to RTL support, change to margin in keeping with the property used in slds-button css.
2. allows the margin value to be declared in the cssString component property: `--fbb-button-spacing: 0.5em;` or more appropriately `--fbb-button-spacing: var(--slds-g-spacing-1)` (or -2) matching the spacing in the SLDS approach to sibling <button> elements.

Happy to discuss whether there should be multiple fallbacks to bake in the standard salesforce spacing e.g.
```
var(--fbb-button-spacing, var(--slds-g-spacing-1, 0.25em))
```
It seems salesforce consistently has slds-g-spacing-1 and then overrides to slds-g-spacing-2


ericrsmith35:
@fromsteinsfdc David - could you please take a look at this and comment?

fromsteinsfdc:
Hey @ericrsmith35 and @readeral, sorry I've been so absentee on this project! This looks great to me.

PR #1657, opened 2025-07-17
feat: Add inlineRichTextEditor LWC

jamesbarker95:
This pull request is for a new Lightning Web Component, `inlineRichTextEditor`, designed for use in Screen Flows. It provides a clean, inline editing experience for rich text, similar to a standard record detail page. Where there is a field with a pencil icon on view then the rich text editor without any buttons upon edit. 

I will send a separate slack message with a recording of the feature as well.

### DIFF
--- a/force-app/main/default/lwc/inlineRichTextEditor/README.md
+++ b/force-app/main/default/lwc/inlineRichTextEditor/README.md
@@ -1,0 +1,27 @@
+# Inline Rich Text Editor
+
+## Description
+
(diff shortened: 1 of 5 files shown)


ericrsmith35:
@jamesbarker95 Are you able to provide a writeup similar to one of these?
https://unofficialsf.com/copy-to-clipboard-button/
https://unofficialsf.com/validation-checker-flow-action/

PR #1662, opened 2025-08-15
dualListBox3: added help text option to the configuration parameters of the screen flow component

abdeslamads:
the `lightning-dual-listbox` component has a parameter to display a help text named `field-level-help`, this option isn't available as a configuration parameter of dualListBox3 screen flow component. I've added it.


ericrsmith35:
I will review this.

PR #1664, opened 2025-08-25
Apex version of Commit Transaction that works with LWR sites.

crouseamplify:
The existing Commit Transaction component is not supported by LWR sites and cannot be replaced by an LWC yet. This Apex action replaces the functionality and works with LWR.


crouseamplify:
Found an issue with this. Closing pull request while I work on it.

Issue #1685, opened 2025-11-21
Implement Lightning Flow Scanner as GitHub Action


RubenHalman:
Hi, thank you for this great package. Given that it includes many flows, I wanted to check if you’d be open to implementing Lightning Flow Scanner as a GitHub Action for automated Salesforce Flow analysis in pull requests.

The scanner analyzes flow metadata for 20+ common issues such as hardcoded IDs, unsafe contexts, inefficient SOQL/DML usage, recursion risks, and missing fault handling. It can be configured as non-blocking and informational only.

PR #1686, opened 2025-11-26
add LFS action to scan flows in PRs

RubenHalman:
Addresses #1685 by implementing LFS as a GitHub Action to scan Salesforce Flows in pull requests.

A few notes on the implementation:

- The action automatically handles deltas.
- By default, it runs all built-in rules. This behavior can be customized by adding a .flow-scanner.yml configuration file.
- Only in SARIF mode does the action block pull requests when there are unaddressed issues, so I chose not to implement blocking behavior outside of that mode.


RubenHalman:
I created a new branch for this

PR #1693, opened 2026-01-30
Scan Salesforce Flows in Pull Requests

RubenHalman:
Addresses #1685 and implements LFS as a GitHub Action to scan Salesforce Flows in Pull Requests.

A few notes on the implementation:

- The action automatically handles deltas.
- By default, it runs all built-in rules. This behavior can be customized by adding a .flow-scanner.yml configuration file.
- Only in SARIF mode does the action block pull requests when there are unaddressed issues, so I chose not to implement blocking behavior outside of that mode.

### DIFF
--- a/.flow-scanner.yml
+++ b/.flow-scanner.yml
@@ -1,0 +1,14 @@
+# Disables layout and documentation related rules for Flow Scanner
+rules:
+  missing-flow-description:
+    enabled: false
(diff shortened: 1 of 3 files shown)


RubenHalman:
@alexed1 @ericrsmith35 have you had a chance to look at this?

